PR #1, opened 2015-07-19
Added a benchmark for Jacobian and subs.

moorepants:
Here is an additional benchmark. It takes about 1s to run so maybe I should make it faster. What do you think?


bjodah:
Looks good to me! I'll merge this and update the benchmarks. We still need to think about how testing should be incorporated in benchmarking, i.e. we only want to include timings which actually produce the correct result. But I don't know the best way to do that yet..

Issue #7, opened 2015-07-21
Are these sympy benchmarks useful?


moorepants:
Maybe we should merge these into this code:

https://github.com/sympy/sympy/tree/master/sympy/benchmarks

bjodah:
Agreed, let's just give them a little more descriptive names perhaps?

PR #10, opened 2015-07-21
[WIP] Initial migration effort of old sympy benchmarks

bjodah:
See #7

### DIFF
--- a/benchmarks/meijerint.py
+++ b/benchmarks/meijerint.py
@@ -58,193 +58,464 @@ def E(expr):
     res2 = integrate(expr*exponential(x, rate)*normal(y, mu1, sigma1),
                      (y, -oo, oo), (x, 0, oo), meijerg=True)
 
-bench = [
-    'MT(x**nu*Heaviside(x - 1), x, s)',
-    'MT(x**nu*Heaviside(1 - x), x, s)',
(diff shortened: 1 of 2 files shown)


bjodah:
One down side with these benchmarks is that they are somewhat contrived, and the sheer number of benchmarks might have a negative effect on brevity of the results..

moorepants:
I haven't evaluated them. I just noticed that we had some benchmarks that have probably not been looked at in a long time. I don't know if they are worth adding. @asmeurer or @certik may know better.

asmeurer:
I'm inclined to believe that the meijerint benchmarks are useful, since they were written by the author of the algorithm.  I know a lot of the benchmarks that are there can stand to be rewritten, though.

asmeurer:
Not sure about the number. Does ASV handle a large number of benchmarks well? It has a feature to only show things that have slowed down significantly, right?

moorepants:
It can run as many benchmarks as you want. You can start by doing a course mesh over the commit history and then a finer mesh or bisection on specific benchmarks and commit ranges to narrow down the offending commits. It can also skip previously failed or successful benchmarks runs.

pv:
The dev version of asv can list benchmarks whose performance regressed, and commit ranges where this occurred. The meijerg integration here might also be more suitable as a parameterized benchmark, since it seems it is only the function to be integrated that varies.

PR #17, opened 2015-07-25
Add benchmarks for logic module

debugger22:
I haven't tested the code because it gives ProcessError(timed out) on my system.

Original benchmarks are here https://github.com/sympy/sympy/tree/master/sympy/logic/benchmarks

@moorepants @bjodah


benchmarks/logic.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class LogicSuite:
+    """
+    Benchmark suite for sympy.logic
+    """
+    def setup(self):
+        self.INPUT = [5 * i for i in range(2, 16)]
+        self.ALGORITHMS = ['dpll', 'dpll2']
+
```
bjodah: Maybe you want to see differences between different algorithms? If so you can introduce `time_dpll(self)` and `time_dpll2(self)`

bjodah:
I get the following error message when running `asv run` on your branch:

```
$ ~/.local/bin/asv run
· Fetching recent changes.
· Creating environments
· No __init__.py file in 'benchmarks/logic-inputs'
```
Same error in asv 0.1.1 and asv master branch.

I see you have quite a few cnf files, do you vision them ever being changed? if not you could put them in a .tar.gz or zip file and have them extracted into a [temporary directory](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp) upon `setup(self)`.

So you employ a SAT solver in logic module, what about optionally employing `pycosat` much like `fastcache` is an optional dependency?

benchmarks/logic.py
debugger22: That would be better.

debugger22:
> I see you have quite a few cnf files, do you vision them ever being changed? if not you could put them in a .tar.gz or zip file and have them extracted into a temporary directory upon setup(self).

I'll do that.

> So you employ a SAT solver in logic module, what about optionally employing pycosat much like fastcache is an optional dependency?

I created an issue https://github.com/sympy/sympy/issues/9728 to discuss.

debugger22:
@bjodah Could you please test if this works? I added a mock `__init__.py` in that directory.

benchmarks/logic.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Benchmark suite for sympy.logic
+    """
+    def setup(self):
+        self.INPUT = [5*i for i in range(2, 16)]
+
+    def time_dpll(self):
+        for test in self.INPUT:
+            file_name = os.path.join(input_path, 'logic-inputs', '%s.cnf' % test)
+            theory = load_file(file_name)
```
bjodah: Maybe all files should be loaded in `setup(self)`? (in the event that loading would be slow)
You could also benchmark the loading a "time_load" which loads a single file.

bjodah:
Great, `asv run` now works as expected.

benchmarks/logic.py
debugger22: +1

asmeurer:
Great. It's actually more important that we be benchmarking functions like `to_cnf`, which are often the real bottleneck.

debugger22:
@bjodah Perhaps this is ready to go.

bjodah:
Yes, look good. Thanks!

debugger22:
Thanks for the review!

PR #18, opened 2015-07-25
Run asv in test suite.

bjodah:
This will give PR authors feedback on whether `asv run` works on their branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ python:
   - "2.7"
 install:
   - pip install -r requirements.txt
-  - pip install asv
+  - pip install asv virtualenv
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -3,8 +3,9 @@ python:
   - "2.7"
 install:
   - pip install -r requirements.txt
-  - pip install pytest
```
debugger22: Why did you remove `pytest`?

debugger22: +1 otherwise.

bjodah:
I removed `pytest` since it's already provided by the travis environment. Tests pass so I am merging this. Thank you for your review.

PR #19, opened 2015-07-28
Check correctness in teardown in integral benchmark.

bjodah:
cf. https://github.com/spacetelescope/asv/issues/296#issuecomment-123493731


moorepants:
This looks fine. Nice solution to the issue. Looks like travis failed though.

PR #24, opened 2016-04-08
Added benchmarks for solving linear matrix systems, Ax=b.

moorepants:


bjodah:
I think GE, LU and ADJ are good candidates for parametrized tests. See e.g. https://github.com/bjodah/sym/blob/master/benchmarks/benchmarks/Lambdify.py#L61for example of syntax.

moorepants:
I switched it. I think it is correct. Not quite sure.

benchmarks/solve.py
```diff
@@ -54,3 +54,31 @@ def time_solve(self):
 
     def time_solve_nocheck(self):
         sympy.solve(self.eqs, *self.p.c, check=False)
+
+
+class TimeMatrixSolve:
+
+    params = ['GE', 'LU' 'ADJ']
```
bjodah: there is a comma missing after LU, making the list: 'GE', 'LUADJ'

benchmarks/solve.py
```diff
@@ -54,3 +54,31 @@ def time_solve(self):
...
+        self.b = sympy.Matrix(n, 1,
+                              lambda i, j: sympy.Symbol('b{}{}'.format(i, j)))
+        self.A_sym = sympy.Matrix(n, n, lambda i, j:
+                                  sympy.Symbol('a{}{}'.format(*sorted((i, j)))))
+
+    def time_solve(self, name):
+
+        self.A.solve(self.b, method=name)
+
+    def time_lusolve(self, name):
```
bjodah: time_lusolve and time_cholesky_solve should be moved to a separate class (since they aren't affected by the parameter)

moorepants: But they use the same setup. Is the setup run once per method or once per all methods in a class?

bjodah: once per method, shared code can be put in a private function in the module, e.g. `def _get_A_b_A_sym()`

moorepants:
@bjodah I've updated this. Merge if you think it is fine.

bjodah:
Great!

PR #25, opened 2016-04-08
Added benchmarks for count_ops and free_symbols.

### LINKED ISSUE #23: Add benchmark for count_ops and free_symbols for large expressions.
One quick way to get some large expressions:

```python
from pydy.model import n_link_pendulum_on_cart
sys = n_link_pendulum_on_cart(5)
rhs = sys.eom_method.rhs()
```
rhs is a 10 x 1 Matrix with huge expressions as the entries.

moorepants:
The setup takes a few seconds now and count_ops does too. I may need to reduce
the expression length.

Fixes #23.

### DIFF
--- a/benchmarks/large_exprs.py
+++ b/benchmarks/large_exprs.py
@@ -36,7 +36,7 @@ class TimeLargeExpressionOperations:
         for expr in self.sym_matrix[:]:
             self.long_expr += expr
 
-        self.super_long_expr = (self.long_expr * self.long_expr).expand()
+        self.super_long_expr = (self.sym_matrix[0, 0] ** 3).expand()
 
     def time_subs(self):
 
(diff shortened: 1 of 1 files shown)


bjodah:
Looks good. I'm inclined to agree that a few seconds sounds a bit too long (how many are a few?).

moorepants:
I shortened the expression so we get:

```
In [1]: from benchmarks.large_exprs import *

In [2]: t = TimeLargeExpressionOperations()

In [3]: %time t.setup()
CPU times: user 1.32 s, sys: 0 ns, total: 1.32 s
Wall time: 1.32 s

In [4]: %time t.time_count_ops()
CPU times: user 80 ms, sys: 0 ns, total: 80 ms
Wall time: 81.6 ms

In [5]: %time t.time_free_symbols()
CPU times: user 16 ms, sys: 0 ns, total: 16 ms
Wall time: 12.5 ms
```

PR #26, opened 2016-04-19
Added benchmark of Sum.doit() related to issue #11030

dibus2:
This is the benchmark with the call to doit on symbolic sum that is much slower on version 1.0 versus old version 0.7.2. Seems like piecewise_fold is where the time is spent.

### DIFF
--- a/benchmarks/sum.py
+++ b/benchmarks/sum.py
@@ -6,6 +6,9 @@ It worked much faster on 0.7.2(3) than on 1.0.
 import sympy
 from sympy import symbols, Function, Sum, I, Symbol, KroneckerDelta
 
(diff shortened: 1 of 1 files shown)


bjodah:
@dibus2 so we are almost there, you just need to create a class called TimeSomethong with a setup and a time_something method, e.g.:

```
class TimeSum:

    def setup(self):
        self.expr = Sum(...)

    def time_doit(self):
        self.expr.doit()
```
save the changes, and run
```
$ git commit -am "Added TimeClass"
$ git push
```

dibus2:
Ok, hope it is better now. 

F.

bjodah:
Great, the `doit()` method ran in 23ms on the CI server.
Thank you for the benchmark, now we will let the benchmark servers run through the commits of sympy.

moorepants:
The Sum.doit benchmark is live: http://www.moorepants.info/misc/sympy-asv/#

PR #30, opened 2016-07-28
Fixed error in setup.

moorepants:
@bjodah I missed this.

### DIFF
--- a/benchmarks/solve.py
+++ b/benchmarks/solve.py
@@ -74,7 +74,7 @@ class TimeMatrixSolve:
 
     def setup(self, name):
 
-        self.A, self.B, _ = _matrix_solve_setup()
+        self.A, self.b, _ = _matrix_solve_setup()
 
     def time_solve(self, name):
 
@@ -85,7 +85,7 @@ class TimeMatrixSolve2:
 
     def setup(self):
 
-        self.A, self.B, self.A_sym = _matrix_solve_setup()
+        self.A, self.b, self.A_sym = _matrix_solve_setup()
 
     def time_lusolve(self):
 
--- a/benchmarks/solve.py
+++ b/benchmarks/solve.py
@@ -83,7 +83,7 @@ class TimeMatrixSolve:
 
 class TimeMatrixSolve2:
 
-    def setup(self, name):
+    def setup(self):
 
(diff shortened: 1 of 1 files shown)


bjodah:
Oh yes, I forgot that even though travis reports passing tests the actual benchmark might have failed, so one has to manually inspect the CI log.... but in this case the reported error in the CI log from the last PR show failing benchmark for `time_kanesmethod_mass_spring_dampe...` want to look into that one as well here?

moorepants:
Just fixed it and tested it locally. To test this we'd need to write version specific tests.

moorepants:
Sorry for merging in the bad code. I hadn't looked at it in a while and didn't test things thoroughly locally first.

bjodah:
No worries. What makes this hard to test automatically is that failed benchmarks does not affect the exit status of `asv` (see [this](https://github.com/spacetelescope/asv/issues/296)), the people behind `asv` were not opposed to adding a flag causing `asv` to exit with failed exit status on failed benchmarks. But I couldn't justify spending time on developing a patch for `asv` at that time (and probably not right now either).

bjodah:
Just saw that someone else wanted this behavior as well: https://github.com/spacetelescope/asv/issues/449

certik:
@bjodah you should upvote that request on the asv repo. ;)

bjodah:
Upvoted, thanks! :)

moorepants:
Note that if we implement that kind of check then it is going to be real pain to write benchmarks that pass for every commit in the project's history. I'd rather they just fail gracefully for old API versions and when the functions don't even exist.

bjodah:
Maybe there could be a flag which we can add to legacy benchmarks along the lines of (note that this is not a real thing):
```
@asv.mark.may_fail
def time_some_old_method():
    ...
```
then we could just mark old benchmarks as "allowed failures" and keep a "strict" setting in our travis config.

moorepants:
I'm merging this. Tested locally and travis is done.

certik:
@moorepants you should upvote that issue https://github.com/spacetelescope/asv/issues/449 as well. They might implement it if they see a lot of interest.

moorepants:
Done

PR #31, opened 2016-07-28
Updated the Kane test so it actually works.

moorepants:
This one had errors too and this should fix that.


moorepants:
The tests are taking forever here. I am going to merge as is, since there are no tests. I want to get this merged and run the benchmarks overnight before I leave work, as I don't have remote access to the machine I use. Hope that's ok.

bjodah:
Yeah I think we can merge at will in this repo generally - it's just benchmarks.

Issue #32, opened 2016-07-28
time_dpll takes 5 seconds to run, this is way too slow for a benchmark


moorepants:
see: http://www.moorepants.info/misc/sympy-asv/#logic.LogicSuite.time_dpll

moorepants:
These take 5s too:
```
[  1.92%] ··· Running solve.TimeSolve01.time_solve                                                                                                                                 5.00s
[  2.00%] ··· Running solve.TimeSolve01.time_solve_nocheck                                                                                                                         5.01s```

bjodah:
I'll see if I can reformulate the solve benchmarks (make the system smaller).
@debugger22 can you do anything about the running time for dpll?

bjodah:
`time_solve` and `time_solve_nocheck` are now both <1s

PR #33, opened 2016-07-30
Make solve benchmark faster (addresses gh-32)

bjodah:
5 seconds per timing is impractical. This should speed it up.


bjodah:
OK, this reduces the running time by a factor 6. Merging.

Issue #36, opened 2017-06-10
Add lambdify benchmark


moorepants:
lambdify (and even the other code printing facilities) could use some benchmarks. See this issue for notes on lambdify: https://github.com/sympy/sympy/issues/12731

PR #37, opened 2017-06-11
Add benchmark of sympy.lambdify (addresses gh-36)

bjodah:
Benchmarks lambdify (both creation ~100 ms) and evaluation (~200 us).


moorepants:
looks good

PR #40, opened 2017-06-27
Parametrize 'modules' kwarg in lambdify benchmark

bjodah:
To partially address: https://github.com/sympy/sympy/issues/12793

~~Don't know why asv isn't picking up on `params`~~

### DIFF
--- a/benchmarks/lambdify.py
+++ b/benchmarks/lambdify.py
@@ -13,18 +13,22 @@ def _get_args_exprs():
 
 class TimeLambdifyCreation:
 
-    def setup(self):
+    params = ((['math'], ['numpy']),)
+
+    def setup(self, modules):
         self.args, self.exprs = _get_args_exprs()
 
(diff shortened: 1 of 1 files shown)


bjodah:
So here are the timing for creating the lambdified function:
```
               =========== =========

                  param1            

               ----------- ---------

                 ['math']   90.74ms 

                ['numpy']   85.93ms 

               =========== =========

```

and evaluating:
```
               =========== ==========

                  param1             

               ----------- ----------

                 ['math']   16.41μs  

                ['numpy']   399.72μs 

               =========== ==========
```
note: we expect performance conscious users to use `numba` or SymEngine.

moorepants:
Why would numpy be significantly slower?

moorepants:
+ 1 to merge

bjodah:
Numpy functions will create numpy objects around Python's floats. So that's the extra overhead.

PR #47, opened 2019-02-23
Add some basic matrix expression benchmarks

asmeurer:


moorepants:
Looks like the Piecewise output may have changed with the latest SymPy releases.

oscargus:
It has. Or rather, there is better simplification of Piecewise now. Including a bug, although I do not know if the bug affects this. See https://github.com/sympy/sympy/pull/15929

oscargus:
The code in https://github.com/sympy/sympy/pull/15929 "works", but I do not have a 2.7 installed so I do not really know how to progress with the 2.7-related problem (except installing a 2.7...). The structure should be improved as well, but once merged the simplified `Piecewise` will most likely change again.

asmeurer:
I didn't realize there were tests here. These benchmarks are unrelated, so I'm merging this.

bjodah:
There were some errors related to this benchmark (shouldn't the methods take self as an argument?):
```
...
TypeError: time_Mul() takes no arguments (1 given)
```

asmeurer:
Oh crap I forgot to push that fix.

PR #50, opened 2019-04-20
Fix travis failing

sylee957:
- Changed the assumptions used in refine
- Deleted print

I think there may be an issue for `refine` not considering `~Ne` as same as `Eq`


bjodah:
Great! Thanks!

PR #45, opened 2018-08-04
Time diagonal eigenval computation

ethankward:
Suggested [here](https://github.com/sympy/sympy/pull/15030).

PR #51, opened 2019-04-20
Benchmark Matrix diagonal eigenvalues

sylee957:
Closes #45

Also done some refactors


asmeurer:
Feel free to just merge stuff into this repo if the tests pass. We need more rigorous tests for this repo (to make sure things work with older SymPy versions for instance), but that can be fixed later.

moorepants:
One thing we'd like to review is the time each new benchmark takes. We'd like to keep the benchmarks as fast as possible since it takes 1 to 2 days to run the suite.

sylee957:
I believe that solving moderate size eigenvalue problems in SymPy can take insanely long hours, especially with symbolic entries.
If this test is causing such problem, we may either have to trim down the size, or only run on the commits after https://github.com/sympy/sympy/pull/15030

moorepants:
Benchmarks should ideally be microsecond operations. Please check timing of all benchmarks before merging.

moorepants:
Also, can we follow the same "don't merge your own PRs without review" in this repo too. Thanks.

bjodah:
We can add inherently slow benchmarks to the `slow_benchmarks/` folder. Those won't be run regularly though...

PR #52, opened 2019-04-21
Improve readme

### LINKED ISSUE #49: Broken link in readme
From https://github.com/sympy/sympy_benchmarks/blob/master/README.rst

http://spacetelescope.github.io/asv is not working
Maybe it should be changed to https://asv.readthedocs.io/en/stable/?

sylee957:
Fixes #49 
https://github.com/sympy/sympy/pull/16652#issuecomment-484289323

We don't need to make a full tutorial here, but I think we need some strong guide for the sympy main developers, about how to run the benchmarks locally to compare across forks or such.

I have added some basic tutorials

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,3 +24,59 @@ When adding benchmarks, mirror the SymPy directory tree for the feature you are
 adding. If your benchmark is not for a specific feature such that creating
 subdirectories for your benchmark is not reasonable, put your benchmark in the
 top level benchmarking directory.
+
+
+Quickstart
+----------
+
+Installation
+^^^^^^^^^^^^
+
+Firstly, you need to install ``asv``
+
+``pip install asv``
+
+or in conda environment, use
+
+``conda install -c conda-forge asv``
+
+And then you may also need to install ``virtualenv``. Because we have
+``"environment_type": "virtualenv"`` in ``asf.conf.json``.
+
+``pip install virtualenv``
+
+or in conda environment, use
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -18,9 +18,65 @@ Results are currently hosted in an ad hoc manner, you may find some results `her
...
+``conda install -c conda-forge asv``
+
+And then you may also need to install ``virtualenv``. Because we have
+``"environment_type": "virtualenv"`` in ``asf.conf.json``.
+
+``pip install virtualenv``
+
+or in conda environment, use
+
+``conda install -c conda-forge virtualenv`` to install the dependency.
```
moorepants: I can't imagine that this is a good idea.

i would be better to recommend using conda environments in asf.conf.json.

bjodah: probably, I can't remember why we opted to keep `virtualenv` as the default.

moorepants: I just swap out the conf file on my benchmark machine and use conda. Here is an issue: https://github.com/sympy/sympy_benchmarks/issues/4

moorepants: And there was this attempt: https://github.com/sympy/sympy_benchmarks/pull/42

bjodah: Ah, now I remember. @sylee957 opened https://github.com/sympy/sympy_benchmarks/pull/54, I added the auto-selection logic to `run_benchmarks.sh`

PR #53, opened 2019-04-21
Reduce matrix size

sylee957:
#51
- Reduce matrix to 5*5
- Make the matrix numeric

It takes 200us on master, and 4ms on sympy 1.1


bjodah:
Great!

Issue #55, opened 2019-04-22
Test benchmarks on travis


sylee957:
We haven't set up much standards for 'good' benchmarks here.
My ideas are

1. Running benchmarks across every major release of SymPy, and check if there is any failing one
2. Check the minimum, and maximum time for the benchmarks in the range.

We may need some scripts for this

PR #56, opened 2019-04-22
Test sympy-1.0+ during CI

bjodah:
To (partially) address #55 
cc @sylee957


bjodah:
we can improve upon this later

PR #58, opened 2019-06-20
Updated README.rst file

ShubhamKJha:
Updated README.rst file to include instructions for running benchmarks locally.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,17 +91,6 @@ If you are comparing the benchmarks of a branch against master, use the commit h
 
 Running benchmarks for a selection of commits
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Generally we're not interested in benchmarking every single commit to the sympy repo. For example, we might be interested in some tagged releases:
-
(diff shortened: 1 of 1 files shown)


bjodah:
thank you for simplifying for review.
I think this looks good, only a minor thing: it looks like the "Running benchmarks for a selection of commits"
section is duplicated now?

ShubhamKJha:
Hi @bjodah how about this python script
```python
from __future__ import absolute_import
import os
from shutil import copy

try:
    from shutil import which as is_available
except ImportError:
    # for Python 2
    from distutils.spawn import find_executable as is_available

target = 'asv.conf.json'
#target = os.path.join(target)
if is_available('conda') is not None:
    config_file = 'asv.conf.conda.json'
else:
    config_file = 'asv.conf.venv.json'

#config_file = os.path.join(config_file)

try:
    copy(config_file, target)
    print("Success")
except:
    print("Error")
```
This would set the `asv.conf.json` (default config file) according to the environment. It will allow users to use `asv` commands directly.

ShubhamKJha:
Closing this due to #57 being merged.

PR #60, opened 2019-07-24
Add matrix getitem benchmarks

sylee957:
As https://github.com/sympy/sympy/issues/17249, some matrix performance can be affected by its internal structure.
This may help track down some unnecessary degradations by changes to the internal array data structure.


moorepants:
This looks fine. Thanks!

PR #61, opened 2019-08-06
added matrix power benchmarks

Pristine-Cat:
Added benchmarks for matrix powers, one case which works well now and should not be broken and two cases which any matrix multiplication or power optimizations should improve.


oscarbenjamin:
Thanks!

PR #64, opened 2020-01-18
Remove uneeded imports for Python 3.8 and SymPy 1.6 compatibility

oscarbenjamin:
The benchmarks don't currently work with Python 3.8 or SymPy master due to imports that no longer exist e.g.:
https://travis-ci.org/sympy/sympy/jobs/638843773?utm_medium=notification&utm_source=github_status

This removes two imports (that are not needed anyway) to fix that.

### DIFF
--- a/benchmarks/lambdify.py
+++ b/benchmarks/lambdify.py
@@ -1,4 +1,3 @@
-from time import clock
 import sympy as sp
 
 def _get_args_exprs():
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
I also had to update the code that used the `Equality.refine` method to use the `refine` function since `Relational` is no longer a subclass of `Expr` and doesn't have a `refine` method any more:
https://github.com/sympy/sympy/pull/18053

PR #65, opened 2020-02-05
Added slow but thorough matrix benchmarks

Pristine-Cat:
These were requested to test the new matrix intermediate simplification in sympy PR #18572.

### DIFF
--- a/slow_benchmarks/matrices.py
+++ b/slow_benchmarks/matrices.py
@@ -1,0 +1,455 @@
+from sympy import Symbol, Matrix, symbols, SparseMatrix, Abs
+from sympy import I, S, simplify
+from sympy.simplify.simplify import simplify
+
+_TEST_SIMPLIFY = False # test simplify after operation?
+
+x, y, z = symbols('x y z')
+
+A = Matrix(8, 8, ([1+x, 1-x]*4 + [1-x, 1+x]*4)*4)
+B = Matrix(8, 8, [x+i for i in range (64)])
+C = Matrix(S('''[
+            [             -3/4,       45/32 - 37*I/16,         1/4 + I/2,      -129/64 - 9*I/64,      1/4 - 5*I/16,      65/128 + 87*I/64,         -9/32 - I/16,      183/256 - 97*I/128],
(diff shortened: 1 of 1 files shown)


slow_benchmarks/matrices.py
```diff
@@ -0,0 +1,455 @@
+from sympy import Symbol, Matrix, symbols, SparseMatrix, Abs
+from sympy import I, S, simplify
+from sympy.simplify.simplify import simplify
+
+_TEST_SIMPLIFY = False # test simplify after operation?
```
oscarbenjamin: Why is this needed?

Pristine-Cat: In the original benchmark since dotprodsimp was optional this was for comparing `func(dotprodsimp=True)` vs. `simplify(func())`. I left the code in just in case at some point someone wants to make the same comparison.

slow_benchmarks/matrices.py
```diff
@@ -0,0 +1,455 @@
...
+DDS = SparseMatrix(DD)
+
+O4 = Matrix(4, 1, [1, 1, 1, 1])
+
+
+class TimePow4:
+    def time_A(self): A**4
+    def time_B(self): B**4
+    def time_C(self): C**4
+    def time_D(self): D**4
```
oscarbenjamin: Maybe could do
```
class _TimeMatrixBase:
    def time_A(self): self.func(A)
    def time_B(self): self.func(B)
    def time_C(self): self.func(C)
    def time_D(self): self.func(D)

class TimePow4(_TimeMatrixBase):
    def func(self, M):
        return M**4
```

Pristine-Cat: Probably, but at this point with my potentially limited time I am focusing on the actual matrix code...

moorepants:
Note that we specifically want to avoid slow benchmarks. I suggest that benchmarks should be under a second here: https://github.com/sympy/sympy_benchmarks/issues/8

Pristine-Cat:
Are the benchmarks in the `slow_benchmarks` directory run by default? I thought I had read than no, so that is where these reside.

moorepants:
I see. I guess those are not run, see https://github.com/sympy/sympy_benchmarks/issues/35.

PR #66, opened 2020-03-17
Added Polygon benchmark

OmarWagih1:
Continuing on the effort in benchmarking sympy, i started by benchmarking the polygon class in the geometry folder.

The benchmarks currently work with asv and i'm looking to benchmark other classes in the geometry module once i get some feedback on this benchmark.

#### Release Notes

* benchmarks
  * Added polygon class benchmarking.

### DIFF
--- a/benchmarks/polygon.py
+++ b/benchmarks/polygon.py
@@ -1,0 +1,72 @@
+from __future__ import print_function, division
+
+from sympy import Rational
+from sympy.geometry import (Line, Point,
+                            Polygon)
+from random import randint
+
+class PolygonTests:
+	def setup(self):
+		self.listOfPolygons = [Polygon(Point(0+w,0+w),
+		Point(3+w,0+w),
+		Point(3+w,3+w),
+		Point(0+w,3+w)) for w in range(10)]
+		self.cutListOfPolygons = [Polygon((-1, -1), (1, Rational(5, 2)), (2, 1), (3, Rational(5, 2)), (4, 2), (5, 3), (-1, 3)) for w in range(10)]
+		self.cutLines = [Line((0, 0), (Rational(9, 2), 3)) for w in range(10)]
+
+	def time_create_polygon(self):
+	    "Creating Polygon"
+	    Polygon(Point(0, 0), Point(3, -1),Point(6, 0), Point(4, 5))
+
+	def polygon_area(self):
+	   "Polygon.area for w in range(10)"
+	   [self.listOfPolygons[w].area for w in range(10)]
+
+	def time_polygon_perimeter(self):
+	    "Polygon.perimeter for w in range(10)"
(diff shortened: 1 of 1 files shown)


sylee957:
I think that some benchmarks are failing. Did you confirm that you can run the example across some previous versions?

OmarWagih1:
Hello, yes some of them are failing for versions 1.0 - 1.4, but that is because these functions were implemented in version 1.5, starting version 1.5 and up, they should work fine.

I assumed even if we had little history of how they benchmark we would want to benchmark them for the future and start somewhere.

OmarWagih1:
> I think that some benchmarks are failing. Did you confirm that you can run the example across some previous versions?

Hello, yes some of them are failing for versions 1.0 - 1.4, but that is because these functions were implemented in version 1.5, starting version 1.5 and up, they should work fine.

I assumed even if we had little history of how they benchmark we would want to benchmark them for the future and start somewhere.

sylee957:
We should look for a way to automatically disable some benchmarks if they are not supported in old versions

OmarWagih1:
> We should look for a way to automatically disable some benchmarks if they are not supported in old versions

The only way i could find in asv to skip tests is to raise a NotImplementedError in the setup class, this is skipped in functions intentionally by ASV (Could refer to this issue https://github.com/airspeed-velocity/asv/issues/638),

so the current way to skip benchmarks is by making them all in separate classes, not ideal coding format as i assume, but the only solution I could find. Benchmarks are now skipped (output is n/a) if the function is not supported.

benchmarks/polygon.py
```diff
@@ -4,69 +4,125 @@
...
-	    [self.listOfPolygons[w].centroid for w in range(10)]
+	    [listOfPolygons[w].centroid for w in range(10)]
 
-	def time_polygon_second_moment(self):
+class PolygonSecondMoment:
+	def setup(self):
+		try:
+			listOfPolygons[0].second_moment_of_area()
+		except AttributeError:
+			raise NotImplementedError
```
smitgajjar: Make sure there are 4 spaces per indentation level, according to PEP8

sylee957:
What about grouping the benchmarks into things that were working persistently across all versions, and things that only works after 1.5?

And I think that you should better use sympy.__version__ than catching the attribute errors, because someone can introduce some bugs for the methods later.

OmarWagih1:
> What about grouping the benchmarks into things that were working persistently across all versions, and things that only works after 1.5?
> 
> And I think that you should better use sympy.**version** than catching the attribute errors, because someone can introduce some bugs for the methods later.

They don't all appear in version 1.5, of the ones that are not in 1.0, there is one that is developed in 1.3 and others that are developed in 1.5, i grouped the attributes together under an umbrella class and the other functions were left as is, I think it would be more descriptive this way.

And yes, that is a very good note, thank you! i used the sympy.__version__ to compare.

PR #68, opened 2020-04-24
Add old benchmarks for solve_lin_sys

oscarbenjamin:
These benchmarks are taken from the main sympy repo where they have been unused. Right now sympy master is too slow for 2 out of the 3 benchmarks so they have been disabled (by raising NotImplementedError in setup()).

Hopefully in the future when sympy master is faster these benchmarks can be enabled conditionally. They run *much* faster under the PR https://github.com/sympy/sympy/pull/18844

The timings in the code comments indicate that at some previous time these benchmarks were a lot faster than they are in current sympy master which suggests that a slowdown was introduced at some point.


oscarbenjamin:
These benchmarks became slow on master in https://github.com/sympy/sympy/pull/12521#issuecomment-619265186

PR #70, opened 2020-04-28
refactoring the core benchmarks and add arit and expand benchmarks

RituRajSingh878:
I am adding basic benchmarks based on the sympy test case and refactor the core benchmarks based on the sympy tests and will add other basic benchmarks in later commits and PRs.
Initially, I was confused about adding the benchmarks as I don't know what I should add in the benchmarks that will be useful, so I am following sympy tests.
If anyone has suggestions for adding benchmarks please give here.

Thanks

### DIFF
--- a/benchmarks/core/Arit.py
+++ b/benchmarks/core/Arit.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from sympy.core import Add, Mul, symbols, I
+from sympy.core import Add, Mul, symbols, I, Pow, S
 
 x, y, z = symbols('x,y,z')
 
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
I think that filenames should be lower case i.e. `arit.py` rather than `Arit.py`.

There are very few lines of code here so I'm not sure there's any need to split across multiple files at this stage. Although I suppose you are doing that to mimic the structure of the sympy codebase which is apparently the intention for the benchmarks...

Does anyone have any opinions about this?

oscarbenjamin:
This will be fine to merge if you change the filename to lower case

RituRajSingh878:
> This will be fine to merge if you change the filename to lower case

I will change the file name and will add more benchmarks to it.

RituRajSingh878:
if it is okay then you can merge it. I will open new PRs for other benchmarks.

oscarbenjamin:
I don't think so many blank lines are needed in test_arit. Otherwise this is good.

oscarbenjamin:
Looks good. Thanks

PR #72, opened 2020-05-04
Add some basic benchmarks for Risch

asmeurer:
There should be better benchmarks for the other algorithms too, but I didn't attempt to do that here. I also didn't attempt to refactor the other integration benchmarks code. 

One question is if these benchmarks are too slow. If they are, we can reduce the `100` parameter.

### DIFF
--- a/benchmarks/integrate.py
+++ b/benchmarks/integrate.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
-import sympy
-from sympy import symbols, Integral, exp
+from sympy import symbols, Integral, exp, log, Add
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
Looks good

asmeurer:
Another benchmark to include, but is currently way too slow, is 

```
risch_integrate(Add(*[log(x + i) for i in range(10)]), x)
```

I think it's partly slow because of the polys (it operates on a Poly with 10 generators), but mostly because the height of the tower and certain ways the algorithm works (every log term is a separate extension in the tower).

PR #73, opened 2020-05-07
add assumption benchmark of core

RituRajSingh878:
I have added assumption benchmarks. I will add a few more commits to the assumption bench.
Then we can merge it. Also, if anything you want to modify in the code, then please tell.

### DIFF
--- a/benchmarks/core/assumption.py
+++ b/benchmarks/core/assumption.py
@@ -1,0 +1,419 @@
+# -*- coding: utf-8 -*-
+
+from sympy import pi, oo
+from sympy.core import Symbol
+from sympy.core import Symbol, S, Rational, Integer
+
+
+class Time_assumption:
+
+    def setup(self):
+        self.x = Symbol('x', real=True, integer=True)
+        self.z_0 = Integer(0)
+        self.z_1 = Integer(1)
+        self.z__1 = Integer(-1)
+        self.oo = S.Infinity
+        self.mm = S.NegativeInfinity
+        self.zoo = S.ComplexInfinity
+        self.nan = S.NaN
+        self.r_pos = Rational(3, 4)
+        self.r_neg = Rational(-3, 4)
+        self.pi = S.Pi
+        self.ex = S.Exp1
+        self.I = S.ImaginaryUnit
+        self.a_1 = Symbol('a', real=False)
+        self.a_2 = Symbol('a', extended_real=False)
+        self.a_3 = Symbol('a', imaginary=True)
+        self.x_0 = Symbol('x', zero=True)
+        self.x_1 = Symbol('x', positive=True)
+        self.x_2 = Symbol('x', nonpositive=True)
+        self.x_3 = Symbol('x', positive=False)
+
+    def time_symbol_unset(self):
+
+        self.x.is_real is True
+        self.x.is_integer is True
+        self.x.is_imaginary is False
+        self.x.is_noninteger is False
+        self.x.is_number is False
+
+
+    def time_zero(self):
+
+	    self.z_0.is_commutative is True
+	    self.z_0.is_integer is True
+	    self.z_0.is_rational is True
+	    self.z_0.is_algebraic is True
+	    self.z_0.is_transcendental is False
+	    self.z_0.is_real is True
+	    self.z_0.is_complex is True
+	    self.z_0.is_noninteger is False
+	    self.z_0.is_irrational is False
+	    self.z_0.is_imaginary is False
+	    self.z_0.is_positive is False
+	    self.z_0.is_negative is False
+	    self.z_0.is_nonpositive is True
+	    self.z_0.is_nonnegative is True
+	    self.z_0.is_even is True
+	    self.z_0.is_odd is False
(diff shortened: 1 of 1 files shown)


benchmarks/core/assumption.py
```diff
@@ -0,0 +1,419 @@
...
+        self.x.is_noninteger is False
+        self.x.is_number is False
+
+
+    def time_zero(self):
+
+	    self.z_0.is_commutative is True
+	    self.z_0.is_integer is True
+	    self.z_0.is_rational is True
+	    self.z_0.is_algebraic is True
```
oscarbenjamin, line 46: I don't think it's necessary to test all of these. Benchmarks should be as narrow as possible (they should time precisely one thing). Also After checking `z_0.is_integer` the results of the other operations will be cached for any Basic instance anyway e.g.:
```julia
In [1]: a = Add(2, 2, evaluate=False)                                                                                                          

In [2]: a._assumptions                                                                                                                         
Out[2]: {}

In [3]: a.is_integer                                                                                                                           
Out[3]: True

In [4]: a._assumptions                                                                                                                         
Out[4]: 
{'integer': True,
 'algebraic': True,
 'rational': True,
 'hermitian': True,
 'commutative': True,
 'irrational': False,
 'complex': True,
 'infinite': False,
 'imaginary': False,
 'transcendental': False,
 'finite': True,
 'extended_real': True,
 'noninteger': False,
 'real': True}
```

oscarbenjamin:
The examples you have picked mostly have precomputed cached assumptions:
```julia
In [5]: S.One._assumptions                                                                                                                     
Out[5]: 
{'commutative': True,
 'real': True,
 'imaginary': False,
 'infinite': False,
 'complex': True,
 'hermitian': True,
 'finite': True,
 'extended_real': True,
 'integer': True,
 'algebraic': True,
 'rational': True,
 'irrational': False,
 'transcendental': False,
 'noninteger': False,
 'zero': False,
 'extended_nonzero': True,
 'nonzero': True,
 'even': False,
 'odd': True,
 'extended_negative': False,
 'negative': False,
 'extended_positive': True,
 'nonpositive': False,
 'extended_nonpositive': False,
 'extended_nonnegative': True,
 'positive': True,
 'nonnegative': True}
```
It's reasonable to time the performance of this but I don't think we would need to do it exhaustively. It would be better to time some things that are slow or have the potential to be slow e.g.:
```julia
In [17]: x = Symbol('x', positive=True)                                                                                                        

In [18]: p = random_poly(x, 10, -10, 10)                                                                                                       

In [19]: p                                                                                                                                     
Out[19]: 
     10      9      8      7      5      4      3      2          
- 4⋅x   + 8⋅x  + 9⋅x  + 3⋅x  + 5⋅x  - 7⋅x  - 6⋅x  - 8⋅x  + 2⋅x + 1

In [20]: %time p.is_positive                                                                                                                   
CPU times: user 396 ms, sys: 1.68 ms, total: 398 ms
Wall time: 403 ms

In [21]: %time p.is_positive                                                                                                                   
CPU times: user 8 µs, sys: 0 ns, total: 8 µs
Wall time: 13.1 µs
```
Note that the second run of is_positive here is instantaneous because the result is cached on the instance:
```julia
In [22]: p._assumptions                                                                                                                        
Out[22]: 
{'positive': None,
 'finite': True,
 'infinite': False,
 'extended_positive': None,
 'extended_real': True,
 'commutative': True,
 'imaginary': False,
 'real': True,
 'complex': True,
 'hermitian': True,
 'extended_nonpositive': None,
 'negative': None,
 'extended_negative': None,
 'zero': None,
 'odd': None,
 'irrational': None,
 'rational': None,
 'algebraic': None,
 'integer': None,
 'extended_nonnegative': None}
```

RituRajSingh878:
Okay, got it. I can close this pr, or may I will keep only one benchmark from each function. But I think it is not necessary to benchmark these assumptions. So whatever suitable you can tell me.

oscarbenjamin:
I think it's good to keep a small number of these benchmarks just to make sure that something like `S.Zero.is_zero` doesn't happen to slow down. I would make a parameterised test that tries a few different values like:
```

class TimeAssumptionsSingleton:
    
    params = ('Half', 'Infinity', 'Exp1', ...)

    def setup(self, name):
        self.singleton = getattr(S, name)

    def time_is_zero(self):
        self.singleton.is_zero

    def time_is_positive(self):
        self.singleton.is_positive
```

asmeurer:
There's also no point to doing the `is False` or `is True`.

RituRajSingh878:
> I think it's good to keep a small number of these benchmarks just to make sure that something like `S.Zero.is_zero` doesn't happen to slow down. I would make a parameterised test that tries a few different values like:
> 
> ```
> 
> class TimeAssumptionsSingleton:
>     
>     params = ('Half', 'Infinity', 'Exp1', ...)
> 
>     def setup(self, name):
>         self.singleton = getattr(S, name)
> 
>     def time_is_zero(self):
>         self.singleton.is_zero
> 
>     def time_is_positive(self):
>         self.singleton.is_positive
> ```

ok, I will try to add like this.

Issue #77, opened 2021-07-24
Travis CI is broken


oscarbenjamin:
Looks like the CI for this repo is broken now.

It should probably be migrated from Travis to GitHub Actions.

See also this sympy issue which discusses the end of free usage for Travis:
https://github.com/sympy/sympy/issues/20374

This PR shows a basic example of setting up actions:
https://github.com/sympy/sympy/pull/20466

SymPy's full actions file can be seen here:
https://github.com/sympy/sympy/blob/master/.github/workflows/runtests.yml

brocksam:
This issue can now be closed with the merging of PR #80.

PR #80, opened 2022-10-11
Migrate CI to GitHub Actions from Travis CI

brocksam:
This PR addresses issue #77, which requests CI be moved to GitHub Actions as there is no longer free usage of Travis CI. In this PR:

- A GitHub Actions workflow (`run-benchmarks.yml`) is added, which is based on the `benchmarks` job in the `runtests.yml` workflow in the main SymPy repo.
- A new asv config file for GitHub Actions is added.
- Deprecated Travis CI-related files are removed.

In the Travis CI job, the benchmarks were run against SymPy `master`, 1.4, 1.3, and 1.2. In the new GitHub Actions workflow, the benchmarks are now only run against SymPy `master` and 1.11, mirroring what is done by the workflow in the main SymPy repo.


.github/workflows/run-benchmarks.yml
```diff
@@ -0,0 +1,40 @@
+name: Run benchmarks
+
+on: [push, pull_request]
+
+jobs:
+
+  benchmarks:
+
+    runs-on: ubuntu-20.04
```
moorepants, line 9:
```suggestion
    runs-on: ubuntu-22.04
```

moorepants:
I guess we have to merge this to trigger the first actions run?

moorepants:
I'm going to do that and you can tidy up things in a new PR if needed.

asmeurer:
Ideally we should run this repo against older SymPy versions that we care about. The idea is that the benchmarks should work against older versions so that we can generate asv plots going back. Really we only need to run the benchmarks once against the old versions, just to make sure they run without error.

brocksam:
> The point of this CI is just to make sure the benchmarks work.

If that's the case (and I agree with this statement) then the CI on this repo is currently doing what it should. It looks like `asv run` has a `--strict` argument that will cause the run command to exit with a non-zero return code if any benchmark fails. If I add this in a new PR and reduce the CI to only run with one SymPy version then that should be suitable. I think we should run against the most recent release rather than `master` as we should be able to guarantee that that works correctly with the benchmarks. Do you agree?

asmeurer:
`master` is important too. If someone adds a new benchmark here we need to make sure it will work with the next version of SymPy too. Ideally we would run every benchmark against every commit, the same as `asv run` does, but that's too much work, so just checking some tags is good enough.

PR #81, opened 2022-10-13
Add additional CSE benchmarks

brocksam:
This PR:

- Introduces a new CSE benchmark based on the "baby example" from Griewank & Walther (2008). This is a simple example function in two variables that requires careful selection of the operation order in the common subexpressions to represent the original expression in its simplest form.
- Introduces a new CSE benchmark based on the "lighthouse example" from Griewank & Walther (2008). This is a simple example matrix function in four variables where one of the two matrix entries is a common subexpression of the other.
- Slightly refactors the existing `time_cse` benchmark.

This PR introduces some `test_` methods in addition to the `time_` methods that will be run by asv. These are included here for bookkeeping with the hope that in the future SymPy's benchmarking will support a wider range of benchmark types (see [this SymPy mailing list thread](https://groups.google.com/g/sympy/c/Txf0hwH96zE) and [this asv GitHub issue](https://github.com/airspeed-velocity/asv/issues/1219)).

### DIFF
--- a/benchmarks/cse.py
+++ b/benchmarks/cse.py
@@ -138,3 +138,94 @@ class GriewankBabyExampleCSE:
     def time_combined_cse(self):
         """Time simultaneous CSE on the baby example and its Jacobian."""
         cse([self.y, self.G])
+
+
+class GriewankLighthouseExampleCSE:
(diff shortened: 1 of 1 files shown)


moorepants:
LGTM

PR #82, opened 2022-10-13
Add differentiation benchmarks

brocksam:
This PR:

- Creates a new benchmark module for benchmarking SymPy's differentiation functions.
- Introduces two new simple differentiation benchmarks based on the "baby example" and the "lighthouse example" from Griewank & Walther (2008).
- Introduces a new large differentiation benchmark based on the Gait2D model from Ackermann and van den Bogert (2010) and Gerritsen et al. (1998), and in [this GitHub repo](https://github.com/csu-hmc/gait2d) by Moore . See also [this related GitHub issue](https://github.com/sympy/sympy/issues/8191).

This PR introduces so `test_` methods in addition to the `time_` methods that will be run by asv. These are included here for bookkeeping with the hope that in the future SymPy's benchmarking will support a wider range of benchmark types (see [this SymPy mailing list thread](https://groups.google.com/g/sympy/c/Txf0hwH96zE) and [this asv GitHub issue](https://github.com/airspeed-velocity/asv/issues/1219)).


moorepants:
One thing about asv is that the benchmarks shouldn't take much time. Does that 2d walking model take long to compute? How to you plan to handle benchmarks that take a long time if so? See #8.

moorepants:
LGTM

PR #83, opened 2023-02-14
Add new CSE benchmark based on Kalman filter matrix expression

brocksam:
Issue https://github.com/sympy/sympy/issues/24696 highlighted that CSE on matrix expressions was extremely slow because terms were eagerly evaluated during the optimisation pass. PR https://github.com/sympy/sympy/issues/24696 implemented a fix that reduces the runtime on this 2x2 example by ~10,000x, even more for larger systems. It's worth adding as a benchmark to avoid performance regressions in the future.

### DIFF
--- a/benchmarks/cse.py
+++ b/benchmarks/cse.py
@@ -237,3 +237,79 @@ class GriewankLighthouseExampleCSE:
     def time_combined_cse(self):
         """Time simultaneous CSE on the lighthouse example and its Jacobian."""
         cse([self.y, self.G])
+
+
+class KalmanFilterMatrixEquationCSE:
+    """Kalman filter example from Matthew Rocklin's SciPy 2013 talk.
+
+    Talk titled: "Matrix Expressions and BLAS/LAPACK; SciPy 2013 Presentation"
+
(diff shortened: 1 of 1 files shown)


brocksam:
Unsure whether this should be treated as a "slow" benchmark or not. Runtime is approximately ~60s. However, https://github.com/sympy/sympy/issues/24696 reduces it to ~6ms. So in the historic case it's "slow", but moving forward it'll be fast.

brocksam:
In that case I think this is good to go from my end then.

moorepants:
Should we wait to merge after your fix goes it? (so it doesn't just add 60 sec to the PR CI)

brocksam:
> Should we wait to merge after your fix goes it? (so it doesn't just add 60 sec to the PR CI)

Yes, should probably wait. https://github.com/sympy/sympy/issues/24696 is ready so and appears approved by reviewers so just waiting for it to be merged, hopefully soon.

asmeurer:
Should this be merged now?

brocksam:
I think so, yes please.

PR #84, opened 2023-03-26
Fixed typos and applied some code conventions

LukasMoll:
Fixed some small typos in the doc strings.
Replaced `object call` -> `_ = object call` to show that it is not being used.
Added some spaces where needed.
Fixed method some method calls `class.method` -> `class.method()`.
Fixed some doctstrings, replaced single quoted with triple quoted doctstring.
Removed unused imports.

Is there an interest in using some linters (e.g. black, pep8 or pylint)?


benchmarks/large_exprs.py
```diff
@@ -74,7 +74,7 @@ def time_cse(self):
 
     def time_free_symbols(self):
 
-        self.super_long_expr.free_symbols
+        self.super_long_expr.free_symbols()
```
moorepants, line 77: `free_symbols` is an attribute not a method.

benchmarks/matrices.py
```diff
@@ -61,30 +61,30 @@ def time_MutableDenseMatrix_getitem(self):
         m = self.M1
         for i in range(m.rows):
             for j in range(m.cols):
-                m[i, j]
+                _ = m[i, j]
```
moorepants, line 64: This seems unnecessary and just makes things more verbose.

benchmarks/polygon.py
```diff
@@ -5,90 +5,90 @@
...
+                          Point(3+w, 0+w),
+                          Point(3+w, 3+w),
+                          Point(0+w, 3+w)) for w in range(10)]
 cutListOfPolygons = [Polygon((-1, -1), (1, Rational(5, 2)), (2, 1), (3, Rational(5, 2)), (4, 2), (5, 3), (-1, 3)) for w in range(10)]
 cutLines = [Line((0, 0), (Rational(9, 2), 3)) for w in range(10)]
 
 class PolygonAttributes:
     def time_create(self):
-        "Creating Polygon"
+        """Creating Polygon"""
```
moorepants, line 17: I think the reason for single quote is so the docstrings don't bog down the output to std. I wouldn't change these.

moorepants:
Thanks. I gave some comments. The various lints are fine, but the other things are unnecessary.

> Is there an interest in using some linters (e.g. black, pep8 or pylint)?

Not much, but we could maybe adopt the same linting that occurs in the main sympy repo.

benchmarks/large_exprs.py
LukasMoll: Sorry, should've checked it better.

benchmarks/matrices.py
LukasMoll: the `_ =` stands for a throwaway value. The purpose is to avoid writing statements that look like they no effect but are supposed to be like that. The `_ =` signifies that the return value is not used.

I'll remove it.

benchmarks/polygon.py
LukasMoll: I'll revert it.

LukasMoll:
> Not much, but we could maybe adopt the same linting that occurs in the main sympy repo.

I'll look into that.

benchmarks/matrices.py
moorepants: These are benchmarks so it is all about timing the execution. So it is fine to call functions, methods, etc. but not use the result (or indicate that we aren't using them).

bjodah:
Thanks!

PR #85, opened 2023-03-26
Add flake8

LukasMoll:
- Added flake8 config (copied from sympy, removed exclude and per-ignore-files)
- Applied flake8 code conventions

### DIFF
--- a/benchmarks/cse.py
+++ b/benchmarks/cse.py
@@ -3,7 +3,10 @@
 from functools import reduce
 from operator import add
 
-from sympy import cse, exp, sin, symbols, tan, Matrix
+import sympy as sym
+from sympy import cse, exp, sin, symbols, tan, Matrix, ImmutableDenseMatrix, MatAdd, MatMul, Transpose, Inverse, \
+    MatrixSymbol
+from sympy.core.singleton import SingletonRegistry
 
 
 def _get_args_exprs(nexprs, nterms):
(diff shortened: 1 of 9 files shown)


moorepants:
This is good. Thanks. You'll need to also enable a flake8 check in the CI.

LukasMoll:
> This is good. Thanks. You'll need to also enable a flake8 check in the CI.

Thanks for the feedback. I copied and modified the runtests.yml file from the sympy project. This contains the instructions only for flake8. I can add the others (ruff, mypy) in the future.

moorepants:
What is the purpose of having mypy? I don't quite see how that is relevant for a set of benchmarks.

LukasMoll:
> What is the purpose of having mypy? I don't quite see how that is relevant for a set of benchmarks.

I think it's nice to have to catch type related bugs and to enforce some code conventions.

Example error it's showing now: 
`benchmarks/solve.py:19: error: Name "expr" already defined on line 7  [no-redef]`
Not sure what the purpose is of `expr = None` but I think I can remove it.

I understand that it can be more annoying then useful, I'll remove mypy.

benchmarks/matrices.py
```diff
@@ -96,10 +96,10 @@ class TimeMatrixPower:
...
+        _ = self.Case1**4
 
     def time_Case2(self):
-        m = self.Case2**4
+        _ = self.Case2**4
 
     def time_Case3(self):
-        m = self.Case3**4
\ No newline at end of file
+        _ = self.Case3**4
```
sylee957, line 105:
```suggestion
        _ = self.Case3**4

```

benchmarks/cse.py
```diff
@@ -239,6 +242,9 @@ def time_combined_cse(self):
         cse([self.y, self.G])
 
 
+S = SingletonRegistry()
```
sylee957, line 245: Maybe this can be imported from sympy

LukasMoll:
> I think that __init__.py from logic-inputs should be removed. That looks like some abuse

Thanks.

Issue #88, opened 2023-06-27
benchmarks for sparse polynomial methods


1e9abhi1e10:
@oscarbenjamin  I want to know whether I create a new file or add a benchmark in the `benchmarks/polys.py` file.
Also, should I add all benchmarks which were I commented on the [Issue](https://github.com/sympy/sympy/pull/25278#issuecomment-1607302917).

oscarbenjamin:
It would be good to have a polynomial benchmarks file if there is not already somewhere that has polynomial benchmarks.

> Also, should I add all benchmarks which were I commented on the

Yes, and there should be many different benchmarks for them like comparing different methods for the same operation e.g. the `prem` function vs the `Poly.prem` method vs the `PolyElement.prem` method etc:
```python
In [1]: f, g = x**2*y + x*z + 1, x*z + y

In [2]: %time prem(f, g, x)
CPU times: user 13.2 ms, sys: 196 µs, total: 13.4 ms
Wall time: 12.3 ms
Out[2]: 
 3      2    2
y  - y⋅z  + z 

In [3]: fp, gp = Poly(f, [x, y, z]), Poly(g, [x, y, z])

In [4]: %time fp.prem(gp)
CPU times: user 3.71 ms, sys: 0 ns, total: 3.71 ms
Wall time: 3.39 ms
Out[4]: Poly(y**3 - y*z**2 + z**2, x, y, z, domain='ZZ')

In [5]: R = ZZ[x,y,z]

In [6]: fpe, gpe = R.from_sympy(f), R.from_sympy(g)

In [7]: %time fpe.prem(gpe)
CPU times: user 1.86 ms, sys: 0 ns, total: 1.86 ms
Wall time: 1.88 ms
Out[7]: y**3 - y*z**2 + z**2
```

1e9abhi1e10:
Now should we also add benchmarks for the `pdiv`, `pquo` and `pexquo`?

oscarbenjamin:
> Moving ahead should we also add benchmarks for the `pdiv`, `pquo` and `pexquo`?

Yes.

oscarbenjamin:
> > Moving ahead should we also add benchmarks for the `pdiv`, `pquo` and `pexquo`?
> 
> Yes.

Actually thinking about this we don't need benchmarks for `pquo`, `pexquo` and `pdiv` because they all use the same code (and that is unlikely to change in future). Just one of these is fine.

SirMarcoss:
[...]

PR #89, opened 2023-06-28
Add benchmarks for pseudo remainder method

1e9abhi1e10:
This PR adds the benchmarks for the pseudo-remainder method.
Related Issue:
#88

See [sympy/sympy#25278](https://github.com/sympy/sympy/pull/25278)

### DIFF
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -20,10 +20,8 @@ class TimePolyManyGens:
 
 class _LinearDenseQuadraticGCD:
     def setup_polys(self, n):
-        self.x, *self.y = symbols("x, y1:9")
-        self.R1 = [self.x] + list(self.y)
-        self.R = ZZ[self.R1]
-        x, y, R = self.x, self.y, self.R
+        x, *y = symbols("x, y1:{}".format(n+1))
+        R = ZZ[[x] + list(y)]
         D = (1 + x + sum(y[:n])) ** 2
         f = D * (-2 + x - sum(y[:n])) ** 2
         g = D * (2 + x + sum(y[:n])) ** 2
@@ -34,10 +32,8 @@ class _LinearDenseQuadraticGCD:
 
 class _SparseGCDHighDegree:
     def setup_polys(self, n):
-        self.x, *self.y = symbols("x, y1:9")
-        self.R1 = [self.x] + list(self.y)
-        self.R = ZZ[self.R1]
-        x, y, R = self.x, self.y, self.R
+        x, *y = symbols("x, y1:{}".format(n+1))
+        R = ZZ[[x] + list(y)]
         D = 1 + x ** (n + 1) + sum([y[i] ** (n + 1) for i in range(n)])
         f = D * (-2 + x ** (n + 1) + sum([y[i] ** (n + 1) for i in range(n)]))
         g = D * (2 + x ** (n + 1) + sum([y[i] ** (n + 1) for i in range(n)]))
@@ -48,10 +44,8 @@ class _SparseGCDHighDegree:
 
 class _QuadraticNonMonicGCD:
     def setup_polys(self, n):
-        self.x, *self.y = symbols("x, y1:9")
-        self.R1 = [self.x] + list(self.y)
-        self.R = ZZ[self.R1]
-        x, y, R = self.x, self.y, self.R
+        x, *y = symbols("x, y1:{}".format(n+1))
+        R = ZZ[[x] + list(y)]
         D = 1 + x ** 2 * y[0] ** 2 + sum([y[i] ** 2 for i in range(1, n)])
         f = D * (-1 + x ** 2 - y[0] ** 2 + sum([y[i] ** 2 for i in range(1, n)]))
         g = D * (2 + x * y[0] + sum(y[1:n])) ** 2
@@ -62,10 +56,8 @@ class _QuadraticNonMonicGCD:
 
 class _SparseNonMonicQuadratic:
     def setup_polys(self, n):
-        self.x, *self.y = symbols("x, y1:9")
-        self.R1 = [self.x] + list(self.y)
-        self.R = ZZ[self.R1]
-        x, y, R = self.x, self.y, self.R
+        x, *y = symbols("x, y1:{}".format(n+1))
+        R = ZZ[[x] + list(y)]
         D = -1 + x * prod(y[:n])
         f = D * (3 + x * prod(y[:n]))
         g = D * (-3 + x * prod(y[:n]))
@@ -75,8 +67,6 @@ class _SparseNonMonicQuadratic:
 
 
 class _TimePREM:
-    """Benchmarks for `prem` method in polynomials."""
-
     def setup(self, n, method):
         (f, g, fp, gp, fpe, gpe, x, y, R) = self.setup_polys(n)
 
@@ -84,28 +74,29 @@ class _TimePREM:
 
         prem_f_g_x = rem(f * LC(g, x) ** (degree(f, x) - degree(g, x) + 1), g, x)
 
-        self.ref_expr = prem_f_g_x
-        self.ref_ring = R(prem_f_g_x.as_expr())
-
         if method == 'prem':
             self.func = lambda: prem(f, g, x)
+            self.expected = prem_f_g_x
 
         elif method == 'poly':
             self.func = lambda: fp.prem(gp)
+            self.expected = prem_f_g_x
 
         elif method == 'sparse':
             self.func = lambda: fpe.prem(gpe)
+            self.expected = R(prem_f_g_x.as_expr())
 
     def teardown(self, n, method):
-        for key, val in self.values.items():
-            if key == 'sparse':
-                if (self.ref_ring - val) != 0:
-                    raise ValueError("Incorrect result, invalid timing:"
-                                        " %s != %s" % (self.ref_ring, val))
+        self.result = self.values[method]  # Get the result for the current method
 
-            elif (self.ref_expr - val).simplify() != 0:
-                raise ValueError("Incorrect result, invalid timing:"
-                                    " %s != %s" % (self.ref_expr, val))
+        if method == 'prem':
+            assert self.result == self.expected
+
+        elif method == 'poly':
+            assert self.result == self.expected
+
+        elif method == 'sparse':
+            assert self.result == self.expected
 
     def time_prem_methods(self, n, method):
         self.values[str(method)] = self.func()
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -1,5 +1,5 @@
 from sympy import symbols, prod
-from sympy.polys import ZZ, Poly, ring
+from sympy.polys import ZZ, Poly
 
 
 class TimePolyManyGens:
@@ -25,10 +25,14 @@ class TimePolyprem:
     params = [(1,), (5,), (10,), (15,)]
 
     def setup(self, size):
-        self.R, self.x, self.y, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10 = ring("x, y, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10", ZZ)
+        self.x, self.y, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10 = symbols("x, y, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10")
         self.y = [self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
         self.values = {}
         self.size = size
+        self.generate_poly_case1()
+        self.generate_poly_case2()
+        self.generate_poly_case3()
+        self.generate_poly_case4()
 
     def generate_poly_case1(self):
         # Linearly dense quartic inputs with quadratic GCDs.
@@ -37,19 +41,16 @@ class TimePolyprem:
         self.f = D * (-2 + self.x - sum(self.y[:n])) ** 2
         self.g = D * (2 + self.x + sum(self.y[:n])) ** 2
 
-    def time_new_prem_case1(self):
-        self.generate_poly_case1()
-        self.values['new_prem_case1'] = prem(self.f, self.g, self.x)
+    def time_prem_case1(self):
+        self.values['prem_case1'] = prem(self.f, self.g, self.x)
 
     def time_Polyprem_case1(self):
-        self.generate_poly_case1()
         n = self.size
         fp, gp = Poly(self.f, self.x, *(self.y)[:n]), Poly(self.g, self.x, *(self.y)[:n])
         self.values['Polyprem_case1'] = fp.prem(gp)
 
     def time_PolyElement_prem_case1(self):
-            self.generate_poly_case1()
-            self.R = ZZ[self.x, self.y, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
+            self.R = ZZ[self.x, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
             fpe, gpe = self.R.from_sympy(self.f), self.R.from_sympy(self.g)
             self.values['PolyElement_prem_case1'] = fpe.prem(gpe)
 
@@ -60,19 +61,16 @@ class TimePolyprem:
         self.f = D * (-2 + self.x ** (n + 1) + sum([self.y[i] ** (n + 1) for i in range(n)]))
         self.g = D * (2 + self.x ** (n + 1) + sum([self.y[i] ** (n + 1) for i in range(n)]))
 
-    def time_new_prem_case2(self):
-        self.generate_poly_case2()
-        self.values['new_prem_case2'] = prem(self.f, self.g, self.x)
+    def time_prem_case2(self):
+        self.values['prem_case2'] = prem(self.f, self.g, self.x)
 
     def time_Polyprem_case2(self):
-        self.generate_poly_case2()
         n = self.size
         fp, gp = Poly(self.f, self.x, *(self.y)[:n]), Poly(self.g, self.x, *(self.y)[:n])
         self.values['Polyprem_case2'] = fp.prem(gp)
 
     def time_PolyElement_prem_case2(self):
-        self.generate_poly_case2()
-        self.R = ZZ[self.x, self.y, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
+        self.R = ZZ[self.x, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
         fpe, gpe = self.R.from_sympy(self.f), self.R.from_sympy(self.g)
         self.values['PolyElement_prem_case2'] = fpe.prem(gpe)
 
@@ -83,19 +81,16 @@ class TimePolyprem:
         self.f = D * (-1 + self.x**2 - self.y[0]**2 + sum([self.y[i]**2 for i in range(1, n)]))
         self.g = D * (2 + self.x * self.y[0] + sum(self.y[1:n]))**2
 
-    def time_new_prem_case3(self):
-        self.generate_poly_case3()
-        self.values['new_prem_case3'] = prem(self.f, self.g, self.x)
+    def time_prem_case3(self):
+        self.values['prem_case3'] = prem(self.f, self.g, self.x)
 
     def time_Polyprem_case3(self):
-        self.generate_poly_case3()
         n = self.size
         fp, gp = Poly(self.f, self.x, *(self.y)[:n]), Poly(self.g, self.x, *(self.y)[:n])
         self.values['Polyprem_case3'] = fp.prem(gp)
 
     def time_PolyElement_prem_case3(self):
-        self.generate_poly_case3()
-        self.R = ZZ[self.x, self.y, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
+        self.R = ZZ[self.x, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
         fpe, gpe = self.R.from_sympy(self.f), self.R.from_sympy(self.g)
         self.values['PolyElement_prem_case3'] = fpe.prem(gpe)
 
@@ -106,18 +101,15 @@ class TimePolyprem:
         self.f = D * (3 + self.x * prod(self.y[:n]))
         self.g = D * (-3 + self.x * prod(self.y[:n]))
 
-    def time_new_prem_case4(self):
-        self.generate_poly_case4()
-        self.values['new_prem_case4'] = prem(self.f, self.g, self.x)
+    def time_prem_case4(self):
+        self.values['prem_case4'] = prem(self.f, self.g, self.x)
 
     def time_Polyprem_case4(self):
-        self.generate_poly_case4()
         n = self.size
         fp, gp = Poly(self.f, self.x, *(self.y)[:n]), Poly(self.g, self.x, *(self.y)[:n])
         self.values['Polyprem_case4'] = fp.prem(gp)
 
     def time_PolyElement_prem_case4(self):
-        self.generate_poly_case4()
-        self.R = ZZ[self.x, self.y, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
+        self.R = ZZ[self.x, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
         fpe, gpe = self.R.from_sympy(self.f), self.R.from_sympy(self.g)
         self.values['PolyElement_prem_case4'] = fpe.prem(gpe)
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -1,5 +1,6 @@
 from sympy import symbols, prod, prem, rem, degree, LC
-from sympy.polys import ZZ, Poly
+from sympy.polys import QQ, Poly
+
 
 class TimePolyManyGens:
     """Time using a Poly with many generators"""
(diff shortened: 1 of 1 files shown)


1e9abhi1e10:
@oscarbenjamin  if anything you want to modify in the code, then please tell me. Also, I little bit confused about the timings that how it should be calculated.

benchmarks/polys.py
```diff
@@ -1,17 +1,146 @@
...
+R = ZZ[x, y, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10]
+
+# Case: Linearly dense quartic inputs with quadratic GCDs
+def bench1_core_prem():
+    for v in range(1, 11):
+        D = (1 + x + sum(y[:v])) ** 2
+        f = D * (-2 + x - sum(y[:v])) ** 2
+        g = D * (2 + x + sum(y[:v])) ** 2
+
+        return prem(f, g, x)
```
oscarbenjamin, line 36: We should time separately for each value of v. This also includes the time to construct the polynomials which it should not.

The benchmarks should be more like this one:
https://github.com/sympy/sympy_benchmarks/blob/3680cc194e5af27663c7e46215bf89789766291e/benchmarks/integrate.py#L46-L66

oscarbenjamin:
> make a separate PR for this?

It could be. The tests won't run on this one until that change is made.

1e9abhi1e10:
there are almost 40 values of F and G from these benchmarks, should I include all ?

oscarbenjamin:
> there are almost 40 values of F and G from these benchmarks, should I include all ?

Maybe just pick some representative ones. It is good to time with smaller examples and also larger examples from each group. Also benchmarks should not take longer than around 1 second so don't use really large examples. We can still include the code to generate the examples of any size but we don't need to have the benchmark suite run all sizes every time it runs.

benchmarks/polys.py
```diff
@@ -1,17 +1,40 @@
...
 
     def time_create_poly(self, n):
-        sympy.Poly(self.x, self.xs)
+        Poly(self.x, self.xs)
 
     def time_is_linear(self, n):
         self.px.is_linear
+
+
+class TimePolyprem:
```
1e9abhi1e10, line 22: @oscarbenjamin Is this method is the right way to add benchmarks?

oscarbenjamin:
Code should be added that can generate pairs of polynomials of different sizes as well as their gcd and what the prem should be. We can reuse this for other benchmarks like gcd later.

The Time class should have a params attribute that says which sizes to use for timings when running the benchmarks suite and the setup method should take a parameter that will be the values from params.

The purpose of storing the results in values is so that the tear down method can check that the result is correct (there is no point reporting timings if a wrong result is computed).

There should be `time_` methods for different `prem` functions.

Many of these things are demonstrated by the example I linked before:
https://github.com/sympy/sympy_benchmarks/blob/3680cc194e5af27663c7e46215bf89789766291e/benchmarks/integrate.py#L46

benchmarks/polys.py
```diff
@@ -1,17 +1,123 @@
...
+    def generate_poly_case1(self):
+        # Linearly dense quartic inputs with quadratic GCDs.
+        n = self.size
+        D = (1 + self.x + sum(self.y[:n])) ** 2
+        self.f = D * (-2 + self.x - sum(self.y[:n])) ** 2
+        self.g = D * (2 + self.x + sum(self.y[:n])) ** 2
+
+    def time_new_prem_case1(self):
+        self.generate_poly_case1()
+        self.values['new_prem_case1'] = prem(self.f, self.g, self.x)
```
oscarbenjamin, line 42: The `generate_poly_case1` method should not be called inside the `time_` method because its time will be included in what is measured by asv. The setup method should be used to prepare the inputs to `prem` and then the `time_` method should do nothing more than the operation that is being timed (`prem`).

benchmarks/polys.py
```diff
@@ -1,17 +1,115 @@
...
+        self.f = D * (3 + self.x * prod(self.y[:n]))
+        self.g = D * (-3 + self.x * prod(self.y[:n]))
+
+    def time_prem_case4(self):
+        self.values['prem_case4'] = prem(self.f, self.g, self.x)
+
+    def time_Polyprem_case4(self):
+        n = self.size
+        fp, gp = Poly(self.f, self.x, *(self.y)[:n]), Poly(self.g, self.x, *(self.y)[:n])
+        self.values['Polyprem_case4'] = fp.prem(gp)
```
oscarbenjamin, line 110: It is important not to put anything in the `time_` method except for the actual thing that we want to time which in this case is the call to `fp.prem`. Anything else will affect the timings recorded by asv. That is why there is a separate `setup` method so that we can setup the objects needed but not include the time for setup in the measured timings.

The `time_prem_case4` method is good because it only times the actual call to `prem`.

The `time_Polyprem_case4` method is not because it first calls `Poly` and then calls `prem`. The creation of the two `Poly` objects should happen in the `setup` method and not in the `time_` method.

oscarbenjamin:
If you look in the CI output under "Run benchmarks" you can find this output which shows the other polys benchmarks being run:
```
[ 42.24%] ··· ======== =============
               param1               
              -------- -------------
                 1       43.9±0.1μs 
                 10     74.9±0.09μs 
                100       699±2μs   
                500     12.5±0.05ms 
              ======== =============

[ 42.41%] ··· polys.TimePolyManyGens.time_is_linear                           ok
[ 42.41%] ··· ======== =============
               param1               
              -------- -------------
                 1      1.69±0.01μs 
                 10      19.9±0.1μs 
                100      569±0.3μs  
                500     12.0±0.01ms 
              ======== =============
```
I don't see any output for the `prem` benchmarks that you have added here so it looks like they are not being run by asv.

Do you see any output for the prem benchmarks when running asv locally?

benchmarks/polys.py
```diff
@@ -1,17 +1,115 @@
...
+
+
+class TimePolyprem:
+    """Benchmark for the prem method"""
+
+    params = [(1,), (5,), (10,), (15,)]
+
+    def setup(self, size):
+        self.x, self.y, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10 = symbols("x, y, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10")
+        self.y = [self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
```
oscarbenjamin, line 29: Why not just
```
self.y = symbols("x, y, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10")
```

benchmarks/polys.py
```diff
@@ -1,17 +1,115 @@
...
+        self.generate_poly_case2()
+        self.generate_poly_case3()
+        self.generate_poly_case4()
+
+    def generate_poly_case1(self):
+        # Linearly dense quartic inputs with quadratic GCDs.
+        n = self.size
+        D = (1 + self.x + sum(self.y[:n])) ** 2
+        self.f = D * (-2 + self.x - sum(self.y[:n])) ** 2
+        self.g = D * (2 + self.x + sum(self.y[:n])) ** 2
```
oscarbenjamin, line 42: I would make this a function separate from the class like:
```python
    def generate_poly_case1(n):
        """Linearly dense quartic inputs with quadratic GCDs."""
        x, *y = syms = symbols(f'x, y{n}')
        D = (1 + x + sum(y)) ** 2
        f = D * (-2 + x - sum(y)) ** 2
        g = D * (2 + x + sum(y)) ** 2
        return f, g, D, syms
```

benchmarks/polys.py
```diff
@@ -1,17 +1,115 @@
...
+
+    def time_Polyprem_case1(self):
+        n = self.size
+        fp, gp = Poly(self.f, self.x, *(self.y)[:n]), Poly(self.g, self.x, *(self.y)[:n])
+        self.values['Polyprem_case1'] = fp.prem(gp)
+
+    def time_PolyElement_prem_case1(self):
+            self.R = ZZ[self.x, self.y1, self.y2, self.y3, self.y4, self.y5, self.y6, self.y7, self.y8, self.y9, self.y10]
+            fpe, gpe = self.R.from_sympy(self.f), self.R.from_sympy(self.g)
+            self.values['PolyElement_prem_case1'] = fpe.prem(gpe)
```
oscarbenjamin, line 55: The indentation is off.

oscarbenjamin:
I'm not sure why the benchmark doesn't seem to run in CI.

1e9abhi1e10:
> Do you see any output for the prem benchmarks when running asv locally?

I am currently experiencing certain challenges while running asv locally, but I am determined to find solutions to overcome them. Do I need to use Python version 3.10 specifically for executing the benchmarks?

oscarbenjamin:
Let's see if the added benchmarks run now.

1e9abhi1e10:
> Let's see if the added benchmarks run now.

Now the benchmarks for the `prem`  method are running, but they are failing due to some reason. I will check it

oscarbenjamin:
You will need to be able to run the benchmarks locally to get it working. Debugging via CI is impossible.

1e9abhi1e10:
@oscarbenjamin Please have a look.

oscarbenjamin:
This increases the total time taken to run the benchmarks from 15 minutes to 25 minutes which is too much. Some of the benchmarks are too slow and should be removed. Generally this benchmark suite cannot include anything that takes longer than a second and really we should limit to around 100ms.

For example time_prem_case1 with n=8 takes 4 seconds and with n=10 takes 17 seconds. Given that asv will run these multiple times in a loop this is too slow for what we can include in the benchmark suite. That means that time_prem_case1 should not be used with n=8 or n=10.

There are different ways to control this but one way is to make separate classes that have different values for params and try different sizes e.g.:
https://github.com/sympy/sympy_benchmarks/blob/3680cc194e5af27663c7e46215bf89789766291e/benchmarks/solve.py#L498-L510

benchmarks/polys.py
```diff
@@ -1,17 +1,124 @@
...
-        sympy.Poly(self.x, self.xs)
+        Poly(self.x, self.xs)
 
     def time_is_linear(self, n):
         self.px.is_linear
+
+
+x, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10 = symbols("x, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10")
+y = [y1, y2, y3, y4, y5, y6, y7, y8, y9, y10]
+R = ZZ[x, y1, y2, y3, y4, y5, y6, y7, y8, y9, y10]
```
oscarbenjamin, line 24: The variables like `y1, y2, ...` are not needed:
```
x, *y = symbols("x, y:10")
R = ZZ[x, *y]
```
These should not be global variables though so move them into the `generate_` methods.

Probably the best thing here is to do like the PyDy examples and have a superclass that can generate examples and subclasses that time specific methods with specific parameter values.

1e9abhi1e10:
Should I add examples of `f` and `g` for specific values of `n`? If yes then will I add around 16 different values for different cases and specific values of `n`.

Like these:
https://github.com/sympy/sympy_benchmarks/blob/3680cc194e5af27663c7e46215bf89789766291e/benchmarks/solve.py#L423-L433

oscarbenjamin:
> Should I add examples of `f` and `g` for specific values of `n`?

No, it is better to keep the code that generates them.

> Like these:

Those are only done like that because the code that generates them needs PyDy which is a separate library. The examples are hard-coded to avoid having the benchmarks depend on other libraries.

benchmarks/polys.py
1e9abhi1e10:
```
x, *y = symbols("x, y:10")
R = ZZ[x, *y]
```
This syntax is only valid in Python 3.11 not in 3.10. For running benchmarks I need python 3.10.


```python

Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from sympy import *     
>>> x, *y = symbols("x, y:10")
>>> R = ZZ[x, *y]
  File "<stdin>", line 1
    R = ZZ[x, *y]
              ^
SyntaxError: invalid syntax

```

benchmarks/polys.py
```diff
@@ -1,17 +1,132 @@
...
+        f = D * (-2 + self.x - sum((self.y)[:n])) ** 2
+        g = D * (2 + self.x + sum((self.y)[:n])) ** 2
+        return f, g
+
+    def generate_poly_case2(self, n):
+        """Sparse GCD and inputs where degree is proportional to the number of variables."""
+        D = 1 + self.x ** (n + 1) + sum([(self.y)[i] ** (n + 1) for i in range(n)])
+        f = D * (-2 + self.x ** (n + 1) + sum([(self.y)[i] ** (n + 1) for i in range(n)]))
+        g = D * (2 + self.x ** (n + 1) + sum([(self.y)[i] ** (n + 1) for i in range(n)]))
+        return f, g
```
oscarbenjamin, line 69: I think that it would be better to make separate classes for each of these cases like:
```python
class LinearDenseQuadraticGCD:
    def generate(self, n):
        ...

class SparseGCDHighDegree:
    def generate(self, n):
        ...
```
Then you could avoid the repetitive code that lists all the cases and the timing output would be a bit more meaningful than just "case1", "case2" etc.:
```
[ 41.24%] ··· ...TimePolyPremFast.time_PolyElement_prem_case1                 ok
[ 41.24%] ··· ======== =============
               param1               
              -------- -------------
                 1      11.7±0.08ms 
                 5       94.5±0.9ms 
                 8        178±2ms   
                 10       186±1ms   
              ======== =============

[ 41.40%] ··· ...TimePolyPremFast.time_PolyElement_prem_case2                 ok
[ 41.40%] ··· ======== ============
               param1              
              -------- ------------
                 1      5.73±0.2ms 
                 5      36.2±0.7ms 
                 8       83.6±2ms  
                 10      101±2ms   
              ======== ============
```

oscarbenjamin:
This is still too slow. It takes the benchmark suite from 15 minutes to 20 minutes. We don't need 25% of the test run to just be spent on prem and we will want to benchmark more methods later.

benchmarks/polys.py
1e9abhi1e10: Should I include the `time_` function for the respective cases in the same class or in different
like this ?

> ```python
> class LinearDenseQuadraticGCD:
>     def generate(self, n):
>
>     def time_(self, n):

oscarbenjamin: There are different ways to do it. The most important thing is how useful the asv output is. The things that we would want are to see how different methods perform relative to each other and how each method performs as n scales up.

Another way is to have pairs of parameters like
```
params = [(1, 1), (1, 2), (1, 5), (2, 1), (2, 2), ...]
```

benchmarks/polys.py
```diff
@@ -1,17 +1,139 @@
...
+
+        self.fp_SparseGCDHighDegree = Poly(self.f_SparseGCDHighDegree, self.x, *(self.y)[:n])
+        self.gp_SparseGCDHighDegree = Poly(self.g_SparseGCDHighDegree, self.x, *(self.y)[:n])
+        self.fpe_SparseGCDHighDegree = self.R.from_sympy(self.f_SparseGCDHighDegree)
+        self.gpe_SparseGCDHighDegree = self.R.from_sympy(self.g_SparseGCDHighDegree)
+
+        self.fp_QuadraticNonMonicGCD = Poly(self.f_QuadraticNonMonicGCD, self.x, *(self.y)[:n])
+        self.gp_QuadraticNonMonicGCD = Poly(self.g_QuadraticNonMonicGCD, self.x, *(self.y)[:n])
+        self.fpe_QuadraticNonMonicGCD = self.R.from_sympy(self.f_QuadraticNonMonicGCD)
+        self.gpe_QuadraticNonMonicGCD = self.R.from_sympy(self.g_QuadraticNonMonicGCD)
```
oscarbenjamin, line 86: This sort of repetitive code is the kind of thing that we should try to avoid.

1e9abhi1e10:
@oscarbenjamin In the next step, should I add benchmarks for the `pdiv` method? Also, should I include those benchmarks in this pull request (PR), or should I create a new one?

benchmarks/polys.py
```diff
@@ -1,17 +1,146 @@
...
+        self.values['Polyprem_QuadraticNonMonicGCD'] = self.fp_QuadraticNonMonicGCD.prem(self.gp_QuadraticNonMonicGCD)
+
+    def time_Polyprem_SparseNonMonicQuadratic(self, n):
+        self.values['Polyprem_SparseNonMonicQuadratic'] = self.fp_SparseNonMonicQuadratic.prem(self.gp_SparseNonMonicQuadratic)
+
+    def time_PolyElement_prem_SparseGCDHighDegree(self, n):
+        self.values['PolyElement_prem_SparseGCDHighDegree'] = self.fpe_SparseGCDHighDegree.prem(self.gpe_SparseGCDHighDegree)
+
+    def time_PolyElement_prem_SparseNonMonicQuadratic(self, n):
+        self.values['PolyElement_prem_SparseNonMonicQuadratic'] = self.fpe_SparseNonMonicQuadratic.prem(self.gpe_SparseNonMonicQuadratic)
```
oscarbenjamin, line 125: The point is to have different subclasses of e.g. `LinearDenseQuadraticGCD` then each subclass would just use
```python
    (self.f, self.g, self.fp, self.gp, self.fpe, self.gpe) = self.generate()
```
It could be something like this:
```python
class _PolyGCDExample:
    def setup(self, n):
        (self.f, self.g, self.fp, self.gp, self.fpe, self.gpe) = self.generate()

class _LinearDenseQuadraticGCD:
    def generate(self, n):
        ...
        return f, g, ...

class _TimePREM(_PolyGCDExample):

    def time_prem(self):
        self.values['prem'] = prem(self.f, self.g)

    def time_prem_PolyElement:
        self.values['prem_PolyElement'] = ...

    def teardown(self, n):
        # here there should be code to check that the answers in self.values are correct.

class TimePREMLinearDenseQuadraticGCD(_LinearDenseQuadraticGCD, _TimePREM):
    params = [1, 3, 5, 8]
```
Alternatively it could be like:
```python
def generate_linear_dense(n):
    ...

def generate_polys(case, n):
    if case == 'linear_dense':
        return generate_linear_dense(n)
    elif case ...

class TimePREM:
    params = [('linear_dense', 1), ('linear_dense', 2), ...]

    def setup(self, case, n):
        self.f, ... = generate_polys(case, n)

    def time_prem(self, case, n):
        ...

    def time_prem_PolyElement(...)
```
There are lots of ways to do this that don't involve the kind of repetitive code I have highlighted here.

We need a structure that makes it easy to add new benchmarks for different functions and different cases without large amounts of repetitive code.

Also different ways of structuring this lead to different output from asv which may be easier or harder to understand. Using multiple parameters leads to output like this:
```
[ 44.11%] ··· solve.TimeMatrixOperations.time_rref                            ok
[ 44.11%] ··· ======== =========== ============ ============
              --                       param2               
              -------- -------------------------------------
               param1       0           2            5      
              ======== =========== ============ ============
                 3      211±0.2μs   27.3±0.1ms   51.4±0.1ms 
                 4      413±0.7μs   68.4±0.2ms   78.6±0.3ms 
              ======== =========== ============ ============
```
If the parameters are not just numbers then that makes it easier to understand. I'm not sure f it is possible to tell `asv` to use something other than just `param1` or `param2` as the parameter names. String parameter values will show in the output though:
```
[ 44.75%] ··· solve.TimeMatrixSolve.time_solve_sym                            ok
[ 44.75%] ··· ======== =========
               param1           
              -------- ---------
                 GE     681±2ms 
                 LU     766±8μs 
                ADJ     202±3ms 
              ======== =========
```
Here we probably want the benchmarks to show how different methods compare for the same class of polynomials. Maybe we want something more like this:
```python
class _LinearDenseQuadraticGCD:
    def generate(self, n):
        ...
        return f, g, ...

class _TimePREM:

    params = [('prem', 1), ('prem', 2), ..., ('prem_PolyElement', 1), ...]

    def setup(self, method, n):
        f, g = self.generate(n)
        if method == 'prem':
            self.func = lambda: prem(f, g)
        elif ...

    def time(self, method, n):
        self.func()

class TimePREMLinearDense(_TimePREM, _LinearDenseQuadraticGCD):
    params = [('prem', 1), ('prem', 2), ..., ('prem_PolyElement', 1), ...]
```
Then for each class of polynomials we will see a table showing how each method scales with `n`.

benchmarks/polys.py
```diff
@@ -1,17 +1,125 @@
...
+                'PolyElement_prem': self.values['PolyElement_prem']
+            }
+
+            computed_res = self.values[method]
+            expected_res = expected_res[method]
+
+            if (computed_res - expected_res).simplify() != 0:
+                raise ValueError("Incorrect result, invalid timing:"
+                                " %s != %s" % (expected_res, computed_res))
+
```
1e9abhi1e10, line 105: Is the right way to implement the `teardown` function?

1e9abhi1e10:
I think the `time_` functions for `prem` method is falling due to `teardown` function.

benchmarks/polys.py
```diff
@@ -1,17 +1,125 @@
...
+                'Poly_prem': self.values['Poly_prem'],
+                'PolyElement_prem': self.values['PolyElement_prem']
+            }
+
+            computed_res = self.values[method]
+            expected_res = expected_res[method]
+
+            if (computed_res - expected_res).simplify() != 0:
+                raise ValueError("Incorrect result, invalid timing:"
+                                " %s != %s" % (expected_res, computed_res))
```
1e9abhi1e10, line 104: Is the right way to implement the `teardown` function?

benchmarks/polys.py
```diff
@@ -1,17 +1,125 @@
...
+
+class TimePREMLinearDenseQuadraticGCD(_LinearDenseQuadraticGCD, _TimePREM):
+    """Calculate time for Linearly dense quartic inputs with quadratic GCDs polynomials."""
+
+    params = [('prem', 'Poly_prem', 'PolyElement_prem'), (1, 3 , 5)] # This case is slow for n=8.
+
+class TimePREMQuadraticNonMonicGCD(_QuadraticNonMonicGCD, _TimePREM):
+    """Calculate time for Quadratic non-monic GCD, F and G have other quadratic factors polynomials."""
+
+    params = [('prem', 'Poly_prem', 'PolyElement_prem'), (1, 3 , 5)] # This case is slow for n=8.
```
oscarbenjamin, line 115: Okay, this looks a lot better.

benchmarks/polys.py
oscarbenjamin: No because this is comparing the computed value with itself.

We either need to have the correct answer stored or we need to check some invariant of the answer. For prem one invariant is:
```python
In [24]: f = x**2 + y**2 - 1

In [25]: g = x + 3*y

In [26]: p = prem(f, g, x)

In [27]: rem(f*LC(g, x)**degree(g, x), g, x) == p
Out[27]: True
```

1e9abhi1e10:
Sorry, I am trying to squash the commits but there is a problem with the extension.
Soon I will fix it!

benchmarks/polys.py
1e9abhi1e10: I think it should be `rem(f*LC(g, x)**(degree(f, x) - degree(g, x) + 1), g, x)` 
Also for  `LC` there are different values for the rings and domains
 ```python
>>> x, y, z = symbols("x y z")
>>> R = ZZ[x, y, z]
>>> gpe = 2*x*y + 2*x*z
>>> gp  = R.from_sympy(gpe)
>>> gp.LC
2
``` 
```python
>>> x, y, z = symbols('x y z')
>>> gpe = 2*x*y + 2*x*z
>>> LC(gpe,  x)
2*y + 2*z
```

The function which can give the desired result is `coeff_wrt(f, x, deg)` which is implemented in https://github.com/sympy/sympy/pull/25278/files

1e9abhi1e10:
Why the asv output is structured in this way?
```
[ 42.11%] ··· ...imePREMSparseGCDHighDegree.time_prem_methods                 ok
[ 42.11%] ··· ================== ======== =============
                    param1        param2               
              ------------------ -------- -------------
                     prem           1        965±20μs  
                     prem           3        1.86±0ms  
                     prem           5      3.38±0.05ms 
                     prem           8      6.31±0.03ms 
                  Poly_prem         1        68.4±2μs  
                  Poly_prem         3      2.63±0.05ms 
                  Poly_prem         5       11.6±0.4ms 
                  Poly_prem         8       45.9±0.6ms 
               PolyElement_prem     1      3.99±0.01ms 
               PolyElement_prem     3      12.1±0.01ms 
               PolyElement_prem     5      26.3±0.05ms 
               PolyElement_prem     8       46.0±0.1ms 
              ================== ======== =============
```

It should be like this
```
[100.00%] ··· polys.TimePREMSparseGCDHighDegree.time_prem_methods                                                                                                                                          ok
[100.00%] ··· ================== ============ ============= ============ =============
              --                                         param2
              ------------------ -----------------------------------------------------
                    param1            1             3            5             8
              ================== ============ ============= ============ =============
                     prem          545±20μs    1.11±0.03ms   2.04±0.2ms   3.76±0.04ms
                  Poly_prem        39.8±3μs    1.66±0.03ms   7.90±0.4ms    29.6±0.8ms
               PolyElement_prem   2.71±0.1ms    7.85±0.3ms    18.2±1ms     31.3±0.9ms
              ================== ============ ============= ============ =============
```
Should I use tuple of a tuple for `params`?

benchmarks/polys.py
oscarbenjamin: > For the failed cases there might be a problem with `LC`

You can convert from one form to another and back if needed e.g.:
```python
In [540]: R(gp.as_expr().as_poly(x).LC())
Out[540]: 2*y + 2*z
```

oscarbenjamin:
> Why the asv output is structured in this way?

Possibly there are too many cases for a horizontal layout. It make a difference to make the names shorter e.g. `PolyElement_prem` could just be `sparse` or something. We don't need `prem` to be part of the name because it is already shown as part of the prem benchmark.

It might be that asv tries to detect the "terminal width" or something. Maybe there is an option to asv to control that.

1e9abhi1e10:
Should this problem be resolved?
```
[ 42.11%] ··· ...imePREMSparseGCDHighDegree.time_prem_methods                 ok
[ 42.11%] ··· ======== ======== =============
               param1   param2               
              -------- -------- -------------
                prem      1        931±5μs   
                prem      3      1.85±0.01ms 
                prem      5      3.20±0.02ms 
                prem      8      6.14±0.05ms 
                poly      1       64.3±0.5μs 
                poly      3      2.47±0.02ms 
                poly      5      11.3±0.03ms 
                poly      8       43.6±0.2ms 
               sparse     1      4.00±0.02ms 
               sparse     3      12.0±0.04ms 
               sparse     5       26.2±0.1ms 
               sparse     8       45.9±0.2ms 
              ======== ======== =============
```

benchmarks/polys.py
```diff
@@ -1,17 +1,148 @@
...
+        self.values = {}
+        self.f = f
+        self.g = g
+        self.fp = fp
+        self.gp = gp
+        self.fpe = fpe
+        self.gpe = gpe
+        self.x = x
+        self.y = y
+        self.R = R
```
oscarbenjamin, line 92: The generate method in one class is assigning things like `self.x` which are also assigned in the `setup_polys` method of the other class. You need to decide whether the generate method is assigning attributes or returning values and there should only be one place where each attribute is assigned.

Remove the `setup_polys` method from this class and just assign all of these attributes in the `generate` method of the other class. In the example code I showed when I referred to `setup_polys` I meant that the `generate` method could be renamed as `setup_polys` not that there should be redundant `setup_polys` method in this class.

oscarbenjamin:
> Should this problem be resolved?

I don't know if it is an easy thing to change but if not then that is fine.

It looks like `param_names` can be used to control the names that are displayed:
https://asv.readthedocs.io/en/stable/benchmarks.html#benchmark-attributes

I don't know if there is anything in the docs about changing the layout.

benchmarks/polys.py
```diff
@@ -1,17 +1,131 @@
...
+
+    def setup(self, method, n):
+        (self.f, self.g, self.fp, self.gp, self.fpe, self.gpe) = self.generate(n)
+        self.values = {}
+        self.x = symbols("x")
+        self.R1 = [self.x] + list(self.y)
+        self.R = ZZ[self.R1]
+
+        self.ref = rem((self.f)*LC(self.g, self.x)**(degree(self.f, self.x) - degree(self.g, self.x) + 1), self.g, self.x) # correct results for prem and poly methods.
+        self.ref_1  = (self.fpe*(self.R(self.gpe.as_expr().as_poly(self.x).LC()))**(self.fpe.degree() - self.gpe.degree() + 1)).rem(self.gpe) # correct result for sparse method.
```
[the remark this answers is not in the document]
1e9abhi1e10: Something like ?
```python
class _SparseNonMonicQuadratic:
    def setup_polys(self, n):
        self.x, *self.y = symbols("x, y1:9")
        self.R1 = [self.x] + list(self.y)
        self.R = ZZ[self.R1]
        x, y, R = self.x, self.y, self.R
        D = -1 + x * prod(y[:n])
        f = D * (3 + x * prod(y[:n]))
        g = D * (-3 + x * prod(y[:n]))
        fp, gp = Poly(f, x, *y[:n]), Poly(g, x, *y[:n])
        fpe, gpe = R.from_sympy(f), R.from_sympy(g)
        return f, g, fp, gp, fpe, gpe, x, y, R


class _TimePREM:
    """Benchmarks for `prem` method in polynomials."""

    def setup(self, method, n):

        (f, g, fp, gp, fpe, gpe, x, y, R) = self.setup_polys(n)

        self.values = {}

        prem_f_g_x = rem(f * LC(g, x) ** (degree(f, x) - degree(g, x) + 1), g, x)

        self.ref_expr = prem_f_g_x
        self.ref_ring = R(prem_f_g_x.as_expr())

        if method == 'prem':
            self.func = lambda: prem(f, g, x)

        elif method == 'poly':
            self.func = lambda: fp.prem(gp)

        elif method == 'sparse':
            self.func = lambda: fpe.prem(gpe)
```

benchmarks/polys.py
```diff
@@ -1,17 +1,148 @@
...
+            if key == 'sparse':
+                if (self.ref_ring - val) != 0:
+                    raise ValueError("Incorrect result, invalid timing:"
+                                        " %s != %s" % (self.ref_ring, val))
+
+            elif (self.ref_expr - val).simplify() != 0:
+                raise ValueError("Incorrect result, invalid timing:"
+                                    " %s != %s" % (self.ref_expr, val))
+
+    def time_prem_methods(self, method, n):
```
1e9abhi1e10, line 123: if I changed the order like this `time_(self, n, method)` and 
`params = [(1, 3 , 5, 8), ('prem', 'poly', 'sparse')]`
 then the asv result will looks like this 
 
 ```
[100.00%] ··· polys.TimePREMSparseNonMonicQuadratic.time_prem_methods                                                                                                                                      ok
[100.00%] ··· ======== ============ ============= ============
              --                        param2
              -------- ---------------------------------------
               param1      prem          poly        sparse
              ======== ============ ============= ============
                 1      850±500μs      48.4±2μs    2.67±0.3ms
                 3      581±100μs    1.46±0.03ms   4.73±0.3ms
                 5       536±20μs     4.09±0.2ms   6.89±0.3ms
                 8      1.11±0.6ms    8.14±0.5ms   7.97±0.4ms
              ======== ============ ============= ============
```
I think the above result would not have any problem with the terminal width.

Comment (oscarbenjamin) [reply benchmarks/polys.py:123]: Yes, that's good. The only thing that would make it better is to set the `param_names` so instead of `param1/2` it could be `degree`  and `implementation` or something like that.

File: benchmarks/polys.py [hunk 18]
```diff
@@ -1,17 +1,137 @@
...
+    def teardown(self, n, method):
+        for key, val in self.values.items():
+            if key == 'sparse':
+                if (self.ref_ring - val) != 0:
+                    raise ValueError("Incorrect result, invalid timing:"
+                                        " %s != %s" % (self.ref_ring, val))
+
+            elif (self.ref_expr - val).simplify() != 0:
+                raise ValueError("Incorrect result, invalid timing:"
+                                    " %s != %s" % (self.ref_expr, val))
```
oscarbenjamin, line 108: The setup method will be called once for each value of `method`. The teardown method will be called also once for each value. Here the teardown method checks *all* values but it should just check the one that was just tested.

It would be better if the setup method just stored the expected result for the given format like:
```
        if method == 'prem':
            self.func = lambda: prem(f, g, x)
            self.expected = prem_f_g_x
```
Then the teardown method can just do:
```
assert self.result == self.expected
```
Also it should not be necessary to use simplify here.

benchmarks/polys.py
```diff
@@ -1,17 +1,137 @@
...
+        self.x, *self.y = symbols("x, y1:9")
+        self.R1 = [self.x] + list(self.y)
+        self.R = ZZ[self.R1]
+        x, y, R = self.x, self.y, self.R
+        D = -1 + x * prod(y[:n])
+        f = D * (3 + x * prod(y[:n]))
+        g = D * (-3 + x * prod(y[:n]))
+        fp, gp = Poly(f, x, *y[:n]), Poly(g, x, *y[:n])
+        fpe, gpe = R.from_sympy(f), R.from_sympy(g)
+        return f, g, fp, gp, fpe, gpe, x, y, R
```
oscarbenjamin, line 74: This function is still mixed up. It is assigning `self.x` but also returning `x`. What is the design here?

When you write a function you need to decide whether the function is supposed to have side effects (like modifying attributes of self) or whether it is supposed to return something. Most functions should just return something and should not have any side effects.

Sometimes (and this is one of those cases) it might be good to have a function that has side effects. In that situation though we should be clear that the intention of the function is to have side effects and so it should typically not return anything.

The assignment to `self.x` above does nothing useful because outside of this function nothing else ever accesses the attribute: it is assigned here and then only accessed here.

benchmarks/polys.py
```diff
@@ -1,17 +1,128 @@
...
+        self.result = self.values[method]  # Get the result for the current method
+
+        if method == 'prem':
+            assert self.result == self.expected
+
+        elif method == 'poly':
+            assert self.result == self.expected
+
+        elif method == 'sparse':
+            assert self.result == self.expected
```
oscarbenjamin, line 99: Think carefully about what the purpose of this `if/elif` chain is...

oscarbenjamin:
This is looking a lot better but the structure is still not quite right for what we want to do when adding more cases later. I reorganised it a bit like this:
```python
from sympy import *

class _GCDExample:
    """A benchmark example with two polynomials and their gcd."""

    def __init__(self, n):
        f, g, d, syms = self.make_poly(n)
        self.f = f
        self.g = g
        self.d = d
        self.x = syms[0]
        self.y = syms[1:]
        self.syms = syms
        self.ring = QQ[syms]

    def make_poly(self, n):
        raise NotImplementedError("Subclass should implement this.")

    def to_poly(self, expr):
        return Poly(expr, self.syms)

    def to_ring(self, expr):
        return self.ring(expr)

    def as_expr(self):
        return (self.f, self.g, self.d, self.syms)

    def as_poly(self):
        return (self.to_poly(self.f), self.to_poly(self.g), self.to_poly(self.d))

    def as_ring(self):
        return (self.to_ring(self.f), self.to_ring(self.g), self.to_ring(self.d), self.ring)


class _SparseGCDHighDegree(_GCDExample):
    """A pair of polynomials in n symbols with a high degree sparse GCD.

    >>> for n in [1, 2, 3, 4, 5]:
    ...     f, g, d, syms = _SparseGCDHighDegree(n).as_expr()
    ...     print(d)
    x**2 + y1**2 + 1
    x**3 + y1**3 + y2**3 + 1
    x**4 + y1**4 + y2**4 + y3**4 + 1
    x**5 + y1**5 + y2**5 + y3**5 + y4**5 + 1
    x**6 + y1**6 + y2**6 + y3**6 + y4**6 + y5**6 + 1
    """

    def make_poly(self, n):
        x, *y = syms = symbols("x, y1:{}".format(n+1))
        d = 1 + x ** (n + 1) + sum([y[i] ** (n + 1) for i in range(n)])
        f = d * (-2 + x ** (n + 1) + sum([y[i] ** (n + 1) for i in range(n)]))
        g = d * (2 + x ** (n + 1) + sum([y[i] ** (n + 1) for i in range(n)]))
        return f, g, d, syms


class _QuadraticNonMonicGCD(_GCDExample):
    """A pair of quadratic polynomials with a non-monic GCD.

    >>> for n in [1, 2, 3, 4, 5]:
    ...     f, g, d, syms = _QuadraticNonMonicGCD(n).as_expr()
    ...     print(d)
    x**2*y1**2 + 1
    x**2*y1**2 + y2**2 + 1
    x**2*y1**2 + y2**2 + y3**2 + 1
    x**2*y1**2 + y2**2 + y3**2 + y4**2 + 1
    x**2*y1**2 + y2**2 + y3**2 + y4**2 + y5**2 + 1
    """
    def make_poly(self, n):
        x, *y = syms = symbols("x, y1:{}".format(n+1))
        d = 1 + x ** 2 * y[0] ** 2 + sum([y[i] ** 2 for i in range(1, n)])
        f = d * (-1 + x ** 2 - y[0] ** 2 + sum([y[i] ** 2 for i in range(1, n)]))
        g = d * (2 + x * y[0] + sum(y[1:n])) ** 2
        return f, g, d, syms


class _TimeOP:
    """
    Benchmarks comparing Poly implementations of a given operation.
    """
    param_names = [
        'size',
        'impl',
    ]

    def setup(self, n, impl):

        examples = self.GCDExampleCLS(n)

        expected = self.expected(*examples.as_expr())

        if impl == 'expr':
            func = self.get_func_expr(*examples.as_expr())
        elif impl == 'dense':
            func = self.get_func_poly(*examples.as_poly())
            expected = examples.to_poly(expected)
        elif impl == 'sparse':
            func = self.get_func_sparse(*examples.as_ring())
            expected = examples.to_ring(expected)

        self.func = func
        self.expected = expected

    def time_op(self, n, impl):
        self.returned = self.func()

    def teardown(self, n, impl):
        assert self.expected == self.returned


class _TimePREM(_TimeOP):

    def expected(self, f, g, d, syms):
        prem_f_g_x = rem(f * LC(g, x) ** (degree(f, x) - degree(g, x) + 1), g, x)
        return prem(f, g, x)

    def get_func_expr(self, f, g, d, syms):
        x = syms[0]
        return lambda: prem(f, g, x)

    def get_func_poly(self, f, g, d):
        return lambda: f.prem(g)

    def get_func_sparse(self, f, g, d, ring):
        return lambda: f.prem(g)


class TimePREM_SparseGCDHighDegree(_TimePREM):
    GCDExampleCLS = _SparseGCDHighDegree
    params = [(1, 3, 5), ('expr', 'dense', 'sparse')]

```
The key point about this is that if we want to test something else like `gcd` or `pquo` etc then it is very easy to add that. We just add a `_TimeGCD` class and everything else carries over for free without duplicating lots of repetitive code:
```python
class _TimeGCD(_TimeOP):

    def expected(self, f, g, d, syms):
        return d

    def get_func_expr(self, f, g, d, syms):
        return lambda: gcd(f, g)

    def get_func_poly(self, f, g, d):
        return lambda: f.gcd(g)

    def get_func_sparse(self, f, g, d, ring):
        return lambda: f.gcd(g)
```
Note how this `_TimeGCD` class contains the minimum possible code that might not be reused for anything other than the specific case of timing GCD across the three implementations. Everything else is factored out.

oscarbenjamin:
So in the end the output looks like:
```
[ 41.78%] ··· polys.TimePREM_LinearDenseQuadraticGCD.time_op                  ok
[ 41.78%] ··· ====== ============= ============= =============
              --                        impl                  
              ------ -----------------------------------------
               size       expr         dense         sparse   
              ====== ============= ============= =============
                1       1.54±0ms    74.6±0.08μs    222±0.5μs  
                3     6.52±0.01ms     1.53±0ms      2.43±0ms  
                5     20.2±0.04ms   8.79±0.04ms   12.4±0.01ms 
              ====== ============= ============= =============

[ 41.95%] ··· polys.TimePREM_QuadraticNonMonicGCD.time_op                     ok
[ 41.95%] ··· ====== ============= ============= =============
              --                        impl                  
              ------ -----------------------------------------
               size       expr         dense         sparse   
              ====== ============= ============= =============
                1       773±2μs      105±0.4μs     411±0.9μs  
                3       3.00±0ms    5.24±0.01ms   5.99±0.02ms 
                5     8.33±0.01ms   10.6±0.03ms   12.9±0.05ms 
              ====== ============= ============= =============

[ 42.11%] ··· polys.TimePREM_SparseGCDHighDegree.time_op                      ok
[ 42.11%] ··· ====== ============= ============= =============
              --                        impl                  
              ------ -----------------------------------------
               size       expr         dense         sparse   
              ====== ============= ============= =============
                1       723±2μs      53.9±0.1μs    151±0.8μs  
                3       1.49±0ms    2.16±0.01ms   2.57±0.01ms 
                5     2.65±0.01ms   9.62±0.02ms   10.9±0.06ms 
                8     5.17±0.01ms   36.6±0.05ms   40.0±0.05ms 
              ====== ============= ============= =============

[ 42.28%] ··· polys.TimePREM_SparseNonMonicQuadratic.time_op                  ok
[ 42.28%] ··· ====== ========= ============= =============
              --                      impl                
              ------ -------------------------------------
               size     expr       dense         sparse   
              ====== ========= ============= =============
                1     467±1μs    62.4±0.1μs    191±0.2μs  
                3     583±1μs     1.79±0ms      2.08±0ms  
                5     693±2μs   4.88±0.01ms   5.30±0.01ms 
                8     858±3μs   9.64±0.03ms   10.2±0.01ms 
              ====== ========= ============= =============
```

oscarbenjamin:
I think this looks good.

Thanks, and let's get it merged.

1e9abhi1e10:
Thanks a lot!

PR #78, opened 2021-08-29
Added boolean algebra benchmark

oscargus:
An initial benchmark to try it out.

PR #91, opened 2023-07-05
[Do Not Merge] Duplicate PR 

1e9abhi1e10:
duplicate PR of #78

### DIFF
--- a/asv.conf.actions.json
+++ b/asv.conf.actions.json
@@ -22,7 +22,7 @@
 
     // This list needs to be updated after each release of sympy. The branch to
     // be checked should always be the previous release.
-    "branches": ["upstream/1.11", "upstream/master", "HEAD"], // for git
+    "branches": ["upstream/1.12", "upstream/master", "HEAD"], // for git
     // "branches": ["tip"],    // for mercurial
 
     // The DVCS being used.  If not set, it will be automatically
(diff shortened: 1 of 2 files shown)


1e9abhi1e10:
@oscarbenjamin  In this PR the added benchmark doesn't running in CI.

oscarbenjamin:
@brocksam should we expect that new benchmarks added in a PR will run in CI for the PR?

I'm not sure exactly what this does:
https://github.com/sympy/sympy_benchmarks/blob/5563b6dd9266d7d9bc87f2c306cc68d7d566a8d4/.github/workflows/run-benchmarks.yml#L29-L34
My guess is that uses the sympy_benchmarks master branch rather than the PR branch.

brocksam:
> @brocksam should we expect that new benchmarks added in a PR will run in CI for the PR?

I agree this is the behaviour that want, but as evidenced above clearly isn't the case right now.

> I'm not sure exactly what this does:
> 
> https://github.com/sympy/sympy_benchmarks/blob/5563b6dd9266d7d9bc87f2c306cc68d7d566a8d4/.github/workflows/run-benchmarks.yml#L29-L34
> 
> My guess is that uses the sympy_benchmarks master branch rather than the PR branch.

Agreed, I've opened PR #92 for this.

PR #93, opened 2023-07-14
Add benchmarks for pseudo quotient method

1e9abhi1e10:
This PR adds the benchmarks for the pseudo-remainder method.
Related Issue:
#88

See [sympy/sympy#25278](https://github.com/sympy/sympy/pull/25278)

### DIFF
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -1,4 +1,4 @@
-from sympy import symbols, prod, prem, rem, degree, LC
+from sympy import symbols, prod, prem, rem, degree, LC, pquo, quo
 from sympy.polys import QQ, Poly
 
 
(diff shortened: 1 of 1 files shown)


benchmarks/polys.py
```diff
@@ -71,7 +70,6 @@ def make_poly(self, n):
         g = d * (2 + x ** (n + 1) + sum([y[i] ** (n + 1) for i in range(n)]))
         return f, g, d, syms
 
-
 class _QuadraticNonMonicGCD(_GCDExample):
```
oscarbenjamin, line 73: Why are these lines deleted? Usually there should be two lines between top level statements like `def` and `class`.

1e9abhi1e10: I thought there should be one line between subclasses, I will correct myself.

1e9abhi1e10:
Closing! 
See https://github.com/sympy/sympy_benchmarks/issues/88#issuecomment-1636092386

oscarbenjamin:
> Closing! See [#88 (comment)](https://github.com/sympy/sympy_benchmarks/issues/88#issuecomment-1636092386)

I meant that we don't need to time all 4 methods. I think that we should have benchmarks for `pdiv` and `prem` but perhaps fewer of them then there are right now for prem.

PR #94, opened 2023-07-16
Add benchmarks for subresultants PRS method

1e9abhi1e10:
Related Issue:

See https://github.com/sympy/sympy/pull/25371

### DIFF
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -48,7 +48,8 @@ class _GCDExample:
         return (self.to_poly(self.f), self.to_poly(self.g), self.to_poly(self.d))
 
     def as_ring(self):
-        return (self.to_ring(self.f), self.to_ring(self.g), self.to_ring(self.d), self.ring)
+        return (self.to_ring(self.f), self.to_ring(self.g), self.to_ring(self.
+        d), self.ring)
 
 
 class _LinearDenseQuadraticGCD(_GCDExample):
@@ -207,7 +208,8 @@ class _TimeOP:
         elif impl == 'sparse':
             func = self.get_func_sparse(*examples.as_ring())
             if isinstance(expected, list):
-                expected = [examples.to_ring(polynomial) for polynomial in expected]
+                expected = [examples.to_ring(polynomial) for polynomial in
+                            expected]
             else:
                 expected = examples.to_ring(expected)
 
@@ -246,7 +248,8 @@ class TimePREM_LinearDenseQuadraticGCD(_TimePREM):
     polynomials."""
 
     GCDExampleCLS = _LinearDenseQuadraticGCD
-    params = [(1, 3, 5), ('expr', 'dense', 'sparse')] # This case is slow for n=8.
+    # This case is slow for n>5.
+    params = [(1, 3, 5), ('expr', 'dense', 'sparse')]
 
 
 class TimePREM_QuadraticNonMonicGCD(_TimePREM):
@@ -256,7 +259,8 @@ class TimePREM_QuadraticNonMonicGCD(_TimePREM):
     """
 
     GCDExampleCLS = _QuadraticNonMonicGCD
-    params = [(1, 3, 5), ('expr', 'dense', 'sparse')] # This case is slow for n=8.
+    # This case is slow for n>5.
+    params = [(1, 3, 5), ('expr', 'dense', 'sparse')]
 
 
 class _TimeSUBRESULTANTS(_TimeOP):
@@ -281,7 +285,8 @@ class _TimeSUBRESULTANTS(_TimeOP):
 
 class TimeSUBRESULTANTS_LinearDenseQuadraticGCD(_TimeSUBRESULTANTS):
     GCDExampleCLS = _LinearDenseQuadraticGCD
-    params = [(1, 2, 3), ('expr', 'dense', 'sparse')] # This case is slow for n>3.
+    # This case is slow for n>3.
+    params = [(1, 2, 3), ('expr', 'dense', 'sparse')]
 
 
 class TimeSUBRESULTANTS_SparseGCDHighDegree(_TimeSUBRESULTANTS):
@@ -291,7 +296,8 @@ class TimeSUBRESULTANTS_SparseGCDHighDegree(_TimeSUBRESULTANTS):
 
 class TimeSUBRESULTANTS_QuadraticNonMonicGCD(_TimeSUBRESULTANTS):
     GCDExampleCLS = _QuadraticNonMonicGCD
-    params = [(1, 2, 3), ('expr', 'dense', 'sparse')] # This case is slow for n>3.
+    # This case is slow for n>3.
+    params = [(1, 2, 3), ('expr', 'dense', 'sparse')]
 
 
 class TimeSUBRESULTANTS_SparseNonMonicQuadratic(_TimeSUBRESULTANTS):
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -1,4 +1,4 @@
-from sympy import symbols, prod, prem, rem, degree, LC, subresultants, resultant
+from sympy import symbols, prod, prem, rem, degree, LC, subresultants
 from sympy.polys import QQ, Poly
 
 
@@ -113,12 +113,20 @@ class _TimeOP:
         if impl == 'expr':
             func = self.get_func_expr(*examples.as_expr())
             expected = examples.to_expr(expected)
+
         elif impl == 'dense':
             func = self.get_func_poly(*examples.as_poly())
-            expected = examples.to_poly(expected)
+            if isinstance(expected, list):
+                expected = [examples.to_poly(polynomial) for polynomial in expected] # for the subresultants type methods the output is in form of list.
+            else:
+                expected = examples.to_poly(expected) # for those methods whose output is only a polynomial not a tuple or list.
+
         elif impl == 'sparse':
             func = self.get_func_sparse(*examples.as_ring())
-            expected = examples.to_ring(expected)
+            if isinstance(expected, list):
+                expected = [examples.to_ring(polynomial) for polynomial in expected]
+            else:
+                expected = examples.to_ring(expected)
 
         self.func = func
         self.expected_result = expected
@@ -173,8 +181,9 @@ class _TimeSUBRESULTANTS(_TimeOP):
 
     def expected(self, f, g, d, syms):
         x = syms[0]
-        subresultant = resultant(f * LC(g, x) ** (degree(f, x) - degree(g, x) + 1), g, x)
-        return [f, g, subresultant]
+        subresultant = subresultants(f, g, x)
+
+        return subresultant
 
     def get_func_expr(self, f, g, d, syms):
         x = syms[0]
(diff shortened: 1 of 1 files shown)


1e9abhi1e10:
The benchmarks for `subresultants` are failing due to the teardown function. I am working on it!

benchmarks/polys.py
```diff
@@ -162,5 +164,44 @@ class TimePREM_QuadraticNonMonicGCD(_TimePREM):
...
+    GCDExampleCLS = _SparseNonMonicQuadratic
+    params = [(1, 3, 5, 8), ('expr', 'dense', 'sparse')]
+
+
+class _TimeSUBRESULTANTS(_TimeOP):
+    """Benchmarks for pseudo-quotient method"""
+
+    def expected(self, f, g, d, syms):
+        x = syms[0]
+        subresultant = resultant(f * LC(g, x) ** (degree(f, x) - degree(g, x) + 1), g, x)
```
1e9abhi1e10, line 176: I think this is the wrong equation

1e9abhi1e10: @oscarbenjamin  Is there any formula that exists to check the correct result?

1e9abhi1e10:
This is the current output
```
[ 42.32%] ··· ...UBRESULTANTS_LinearDenseQuadraticGCD.time_op                 ok
[ 42.32%] ··· ====== ============= ============= =============
              --                        impl                  
              ------ -----------------------------------------
               size       expr         dense         sparse   
              ====== ============= ============= =============
                1     2.07±0.01ms     541±2μs       1.42±0ms  
                2     4.62±0.01ms   3.17±0.01ms     7.01±0ms  
                3       9.93±0ms    18.1±0.04ms   31.2±0.04ms 
              ====== ============= ============= =============

[ 42.48%] ··· ...meSUBRESULTANTS_QuadraticNonMonicGCD.time_op                 ok
[ 42.48%] ··· ====== ============= ============= =============
              --                        impl                  
              ------ -----------------------------------------
               size       expr         dense         sparse   
              ====== ============= ============= =============
                1     1.24±0.01ms     1.21±0ms    7.82±0.01ms 
                2       3.78±0ms    7.28±0.01ms   18.0±0.04ms 
                3     18.8±0.05ms    157±0.2ms     248±0.3ms  
              ====== ============= ============= =============

[ 42.65%] ··· ...imeSUBRESULTANTS_SparseGCDHighDegree.time_op                 ok
[ 42.65%] ··· ====== ========== ============= =============
              --                      impl                 
              ------ --------------------------------------
               size     expr        dense         sparse   
              ====== ========== ============= =============
                1     970±4μs     253±0.4μs     497±0.9μs  
                2     1.32±0ms     1.42±0ms    2.03±0.01ms 
                3     1.84±0ms   6.59±0.02ms   7.81±0.01ms 
                5     3.20±0ms   30.8±0.02ms    34.1±0.1ms 
              ====== ========== ============= =============

[ 42.81%] ··· ...UBRESULTANTS_SparseNonMonicQuadratic.time_op                 ok
[ 42.81%] ··· ====== ========= ============= =============
              --                      impl                
              ------ -------------------------------------
               size     expr       dense         sparse   
              ====== ========= ============= =============
                1     632±1μs     255±1μs      673±0.6μs  
                2     701±2μs    598±0.9μs      1.11±0ms  
                3     759±1μs   7.33±0.02ms   8.05±0.02ms 
                5     885±2μs   20.3±0.05ms   21.2±0.02ms 
              ====== ========= ============= =============
```

benchmarks/polys.py
oscarbenjamin: I don't know if there is a formula. Checking against `subresultants` is fine though.

oscarbenjamin:
I think that this looks good but I think that we should reduce the total number of benchmarks that are reported for both PREM and subresultants.

You can see what it currently looks like when benchmark results are reported to a PR here:
https://github.com/sympy/sympy/pull/25371#issuecomment-1636901713

This PR would double that. Now that we can see these examples and how the benchmarks compare the question is: which of these different benchmarks is potentially reporting something interesting?

For GCD probably all cases are interesting (they were designed for GCD) but for PREM it does not seem that e.g. the `TimePREM_SparseNonMonicQuadratic` timings demonstrate something that is particularly informative compared to what the `TimePREM_SparseGCDHighDegree` timings show.

oscarbenjamin:
As an aside the timings here comparing `expr`, `dense` and `sparse` demonstrate a common problem with the sympy benchmark suite. The `expr` timings suggest that it is faster than both `dense` and `sparse` in most cases (up to 20x faster for larger problems). This is almost certainly not true though: under the hood the `expr` version of e.g. `prem` uses either the `dense` or `sparse` implementation so it is necessarily slower.

The reason for this is that many SymPy operations (with `expr`) use a cache and the benchmarks report timings from repeating an operation many times while the cache is in operation. It means that the timings reported for `expr` are often only really representative of the performance of the cache rather than the actual time that it takes to compute something that is not already cached.

1e9abhi1e10:
> I think that this looks good but I think that we should reduce the total number of benchmarks that are reported for both PREM and subresultants.

The only difference in the timing of `prem` and `subresultants` benchmarks is `prem` uses less time and `subresultants` uses more time.
I didn't see any case where it contradicts it just add some `ms` in subresultants.
I think we should keep the `subresultants` benchmarks.

1e9abhi1e10:
> As an aside the timings here comparing `expr`, `dense` and `sparse` demonstrate a common problem with the sympy benchmark suite. The `expr` timings suggest that it is faster than both `dense` and `sparse` in most cases (up to 20x faster for larger problems). This is almost certainly not true though: under the hood the `expr` version of e.g. `prem` uses either the `dense` or `sparse` implementation so it is necessarily slower.
> 
> The reason for this is that many SymPy operations (with `expr`) use a cache and the benchmarks report timings from repeating an operation many times while the cache is in operation. It means that the timings reported for `expr` are often only really representative of the performance of the cache rather than the actual time that it takes to compute something that is not already cached.

Do you mean we should remove the `expr` ?

oscarbenjamin:
> Should we remove the `expr` or is there any method so that we can able to calculate the actual timing of `expr`?

No, we can leave it there. We just need to recognise that those benchmarks are not always representative of any actual improvement or regression in performance.

oscarbenjamin:
> `subresultants` alternatively use `prem` , so I don't think that we should keep `prem` benchmarks
> It's good for us to keep the benchmarks of `subresultants` and `gcd`.

I would like to keep something for PREM because perhaps in future it would not be used by subresultants for example. I just don't think we need so many benchmarks for PREM. Also if the different cases are not particularly interesting for revealing different aspects of the performance of subresultants then we don't need so many cases for that either.

1e9abhi1e10:
> > `subresultants` alternatively use `prem` , so I don't think that we should keep `prem` benchmarks
> > It's good for us to keep the benchmarks of `subresultants` and `gcd`.
> 
> I would like to keep something for PREM because perhaps in future it would not be used by subresultants for example. I just don't think we need so many benchmarks for PREM. Also if the different cases are not particularly interesting for revealing different aspects of the performance of subresultants then we don't need so many cases for that either.

We should keep `TimePREM_LinearDenseQuadraticGCD` and `TimePREM_QuadraticNonMonicGCD` for PREM and 
`TimeSUBRESULTANTS_LinearDenseQuadraticGCD` , `TimeSUBRESULTANTS_QuadraticNonMonicGCD` and `TimeSUBRESULTANTS_SparseNonMonicQuadratic` for subresultants ?

oscarbenjamin:
> We should keep `TimePREM_LinearDenseQuadraticGCD` and `TimePREM_QuadraticNonMonicGCD` for PREM and
> `TimeSUBRESULTANTS_LinearDenseQuadraticGCD` , `TimeSUBRESULTANTS_QuadraticNonMonicGCD` and `TimeSUBRESULTANTS_SparseNonMonicQuadratic` for subresultants ?

Why in particular those benchmarks?

Or is that just a random choice?

Do `TimePREM_LinearDenseQuadraticGCD` and `TimePREM_QuadraticNonMonicGCD` demonstrate different aspects of the performance of `prem` somehow?

Are the other cases less relevant than those two?

1e9abhi1e10:
For `TimePREM_SparseGCDHighDegree` the polynomials have a high degree of GCD sparse. While a degree is high,  sparsity might make the pseudo remainder method less efficient compared to other approaches like other modular methods.

1e9abhi1e10:
We can also reduce the values of `size` or `n` 
`params = [(1, 3, 5, 8), ('expr', 'dense', 'sparse')`, maybe the two values will be enough.

oscarbenjamin:
> For `TimePREM_SparseGCDHighDegree` the polynomials have a high degree of GCD sparse. While a degree is high, sparsity might make the pseudo remainder method less efficient compared to other approaches like other modular methods.

Okay, well that seems like a good reason. Maybe some docstrings should be added to the benchmark cases explaining why they are chosen. Also it would be good to add docstrings to the examples types like `_LinearDenseQuadraticGCD` so that we can have some idea when looking at the code what the polynomials they generate will look like.

I put some docstrings in the suggested code at https://github.com/sympy/sympy_benchmarks/pull/89#issuecomment-1632857681 as an illustration of how that might look. Perhaps it would just be good to pick a particular `n` and show what `f`, `g` and `d` would be for that case explaining how `n` relates to the generated polynomials.

1e9abhi1e10:
@oscarbenjamin Please have a look!

benchmarks/polys.py
```diff
@@ -111,12 +193,20 @@ def setup(self, n, impl):
...
             func = self.get_func_expr(*examples.as_expr())
             expected = examples.to_expr(expected)
+
         elif impl == 'dense':
             func = self.get_func_poly(*examples.as_poly())
-            expected = examples.to_poly(expected)
+            if isinstance(expected, list):
+                expected = [examples.to_poly(polynomial) for polynomial in expected] # for the subresultants type methods the output is in form of list.
+            else:
+                expected = examples.to_poly(expected) # for those methods whose output is only a polynomial not a tuple or list.
```
oscarbenjamin, line 202: I have mentioned this a few times now:

Lines of code should not be this long. It is especially pointless to make the line too long by adding a long comment at the end of the line: the comment can go anywhere so put it on a different line.
```python
          # for those methods whose output is only a polynomial
          expected = examples.to_poly(expected) 
```

You might think that this looks reasonable if you have your editor in fullscreen mode but most people do not do that. Usually I like to view code side by side with other things (e.g. a browser or a document or something) so I do not fullscreen the code window.

1e9abhi1e10: I apologize for the same mistake again.
I think in that there is no need to wrap the line for the comment. But I should for sure take care of this next time

benchmarks/polys.py
```diff
@@ -111,12 +194,24 @@ def setup(self, n, impl):
...
         elif impl == 'dense':
             func = self.get_func_poly(*examples.as_poly())
-            expected = examples.to_poly(expected)
+            if isinstance(expected, list):
+            # for the subresultants type methods the output is in form of list.
+                expected = [examples.to_poly(polynomial) for polynomial in
+                            expected]
+            else:
+            # for those methods whose output is only a polynomial.
+                expected = examples.to_poly(expected)
```
oscarbenjamin, line 206: The formatting here is weird. Usually a comment should line up with the line below. Also I would not break a line in a list comprehension. Either allow the line to be a little longer than usual or shorten the code somehow:
```python
            if isinstance(expected, list):
                # some methods output a list of polynomials
                expected = [examples.to_poly(p) for p in expected]
            else:
                # others output only a single polynomial.
                expected = examples.to_poly(expected)
```

oscarbenjamin:
Okay, looks good.

1e9abhi1e10:
thanks!

PR #95, opened 2023-08-10
Add benchmarks for gcd methods

1e9abhi1e10:
Related Issue:

See https://github.com/sympy/sympy/pull/25371

### DIFF
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -1,4 +1,4 @@
-from sympy import symbols, prod, prem, rem, degree, LC, subresultants
+from sympy import symbols, prod, prem, rem, degree, LC, subresultants, gcd
 from sympy.polys import QQ, Poly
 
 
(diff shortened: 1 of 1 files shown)


1e9abhi1e10:
Should I add these benchmarks for only `dense` and `expr` cases ?
```
x  = symbols("x")
f = x**2 + I
Traceback (most recent call last):
  File "<pyshell#22>", line 1, in <module>
    f = x**2 + I
TypeError: unsupported operand type(s) for +: 'PolyElement' and 'ImaginaryUnit'
```

oscarbenjamin:
> Should I add these benchmarks for only `dense` and `expr` cases ?

We should test the sparse polynomials as well:
```python
In [1]: p1 = ZZ_I[x](x + I)

In [2]: p2 = ZZ_I[x](x - I)

In [3]: (p1*p2).gcd(p2)
Out[3]: x + (0 + -1*I)

In [4]: type(_)
Out[4]: sympy.polys.rings.PolyElement
```

PR #97, opened 2023-08-26
Add benchmarks for GCD when input is Gaussian Integer

1e9abhi1e10:
Related Issue:

See https://github.com/sympy/sympy/pull/25442

### DIFF
--- a/benchmarks/polys.py
+++ b/benchmarks/polys.py
@@ -1,4 +1,4 @@
-from sympy import symbols, prod, prem, rem, degree, LC, subresultants, gcd
+from sympy import symbols, prod, prem, rem, degree, LC, subresultants, gcd, I
 from sympy.polys import QQ, Poly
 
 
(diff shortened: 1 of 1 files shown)


1e9abhi1e10:
@oscarbenjamin  These errors occur, Do We need to change these?
```
asv run --config asv.conf.actions.json
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/[3](https://github.com/sympy/sympy_benchmarks/actions/runs/5982262043/job/16231134344?pr=97#step:6:3).10.12/x6[4](https://github.com/sympy/sympy_benchmarks/actions/runs/5982262043/job/16231134344?pr=97#step:6:4)
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.10.12/x[6](https://github.com/sympy/sympy_benchmarks/actions/runs/5982262043/job/16231134344?pr=97#step:6:6)4/lib
Couldn't load asv.plugins.mamba because
No module named 'mamba'
Couldn't load asv.plugins.virtualenv because
No module named 'packaging'
· Unknown environment type 'virtualenv'
· No environments selected
Error: Process completed with exit code 1.
```

oscarbenjamin:
> These errors occurs, Do We need to change these?

This is because of the new asv release. I'm not sure if this is fixed now.

oscarbenjamin:
Looks like it still fails.

benchmarks/polys.py
```diff
@@ -341,4 +371,24 @@ class TimeGCD_QuadraticNonMonicGCD(_TimeGCD):
...
+    def get_func_expr(self, f, g, d, syms):
+        return lambda: gcd(f, g)
+
+    def get_func_poly(self, f, g, d):
+        return lambda: f.gcd(g)
+
+
+class TimeGCD_GaussInt(_TimeGaussianInt):
+    GCDExampleCLS = _GaussianInteger
+    params = [(1, 2, 3), ('expr', 'dense')]
```
oscarbenjamin, line 394: This should also time the sparse implementation because that is what is going to be changed.

1e9abhi1e10: I added the sparse method but it is failing. Should I make the separate class and other functions for `Gaussian` inputs

oscarbenjamin: The code should be fixed so that it works. The `to_ring` method assumes that the domain is always `QQ` which will fail for Gaussian integers.

benchmarks/polys.py
```diff
@@ -39,7 +39,10 @@ def to_poly(self, expr):
         return Poly(expr, self.syms)
 
     def to_ring(self, expr):
-        return self.ring(expr)
+        try:
+            return self.ring(expr)
+        except:
+            return expr
```
oscarbenjamin, line 45: Bare except should never be used.

Anyway this is fixing the problem in the wrong place.

benchmarks/polys.py
oscarbenjamin: What needs to be changed is the faulty assumption that the domain is `QQ` here:
https://github.com/sympy/sympy_benchmarks/blob/e4f235f2d890d99024f90e96a3ab83af9bb51760/benchmarks/polys.py#L33
The domain should not be hard-coded as `QQ`.

benchmarks/polys.py
```diff
@@ -195,7 +226,10 @@ def setup(self, n, impl):
             expected = examples.to_expr(expected)
 
         elif impl == 'dense':
-            func = self.get_func_poly(*examples.as_poly())
+            if examples.ring.domain == ZZ_I:
+                func = self.get_func_poly(*examples.as_expr())
+            else:
+                func = self.get_func_poly(*examples.as_poly())
```
oscarbenjamin, line 232: It should not be necessary to check the domain here like this.

benchmarks/polys.py
```diff
@@ -341,4 +387,30 @@ class TimeGCD_QuadraticNonMonicGCD(_TimeGCD):
...
+
+    def get_func_poly(self, f, g, d, syms):
+        fp = Poly(f)
+        gp = Poly(g)
+        return lambda: fp.gcd(gp)
+
+    def get_func_sparse(self, f, g, d, ring):
+        fpe = ZZ_I[ring](f)
+        gpe = ZZ_I[ring](g)
+        return lambda: fpe.gcd(gpe)
```
oscarbenjamin, line 412: These conversions should not be here.

oscarbenjamin:
The domain should just be returned by the `make_poly` function like:
```
return f, g, d, syms, ZZ
```

benchmarks/polys.py
1e9abhi1e10: If we didn't check the domain then how can we exclude the statement `func = self.get_func_poly(*examples.as_ring)`

benchmarks/polys.py
```diff
@@ -341,4 +373,28 @@ class TimeGCD_QuadraticNonMonicGCD(_TimeGCD):
 
 class TimeGCD_SparseNonMonicQuadratic(_TimeGCD):
     GCDExampleCLS = _SparseNonMonicQuadratic
-    params = [(1, 3, 5), ('expr', 'dense', 'sparse')]
\ No newline at end of file
+    params = [(1, 3, 5), ('expr', 'dense', 'sparse')]
+
+
+class _TimeGaussianInt(_TimeOP):
+    """Benchmarks for GCDs method when input is Gaussian Integer"""
```
oscarbenjamin, line 380: This class just duplicates `_TimeGCD` so it is unnecessary.

1e9abhi1e10: Removed it.

benchmarks/polys.py
```diff
@@ -180,32 +180,15 @@ class _GaussianInteger(_GCDExample):
...
-            I*(-y4 + y5) + y1*(y2 + 1) + -y1*y2 - y1*y3 - y1*(y3 + 1) -
-            I*y5 + -I*y6 + -y1*(y2 + 1) - y1*y3 - y1*(y3 + 1) + I*y5 - 3*y1*y2
-            + y1*(y3 + 1) + y1*(y3 + 1) + -y1*y2 - y1*(y2 + 1) - y1*y3 +
-            y1*y2 + y1*y3 + y1*(y3 + 1) + y1*(y2 + 1) + y1*y3)*d)
+        f = (-I*x**n - 2*y1*y3 - I*(-y2 + n) - I*(y4 + y5))*d
 
-        return f, g, d, syms, ZZ_I
+        g = (-I*x**(n + 1) - 2*y1*(y3 + n) - I*(-y2 + n + 1) - I*(y4 + y5))*d
 
+        return f, g, d, syms, ZZ_I
```
oscarbenjamin, line 191: Why has the example been changed?

oscarbenjamin:
Why is `x` being passed here:
https://github.com/sympy/sympy_benchmarks/blob/6aac3803151a3b667224a52d2be93a60030dc91c/benchmarks/polys.py#L325

benchmarks/polys.py
1e9abhi1e10: In the revised example, I identified duplicate terms, and when I simplified the expression, it resulted in a shorter and more concise form

PR #99, opened 2024-01-20
Enable JSONC Handling for .json Files in Git

Ravindu-Hirimuthugoda:
Fix: https://github.com/sympy/sympy_benchmarks/issues/96

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.json linguist-language=jsonc


asmeurer:
I'm not sure github actually reads gitattributes (see https://github.com/Ravindu-Hirimuthugoda/sympy_benchmarks/blob/master/asv.conf.conda.slow.json), but this seems like an OK change regardless.

PR #102, opened 2024-11-12
Add benchmark of Add

bjodah:
From:
https://github.com/sympy/sympy/pull/27254

```
[50.00%] ··· core.arit.Time_Arit.time_Add_thousands       11.4±0ms
```


benchmarks/core/arit.py
```diff
@@ -27,6 +27,12 @@ def time_Add_xy(self):
     def time_Add_xyz(self):
         Add(*[x, y, z])
 
+    def time_Add_thousands(self):
+        # from https://github.com/sympy/sympy/pull/27254
+        a = symbols("a0:2000")
+        b = Add(*a);
```
isuruf, line 33: Thanks for adding this @bjodah. Shall we move these two lines to the top so that it doesn't end up in the timing results?

bjodah: yes, sounds good.

bjodah:
so this now takes about 9 ms on my machine.

PR #105, opened 2025-12-12
Add dm extension benchmark

ASaha-os:
Add DomainMatrix extension field benchmark

Add performance benchmark for DomainMatrix.to_DM(extension=True) to measure the impact of FLINT vs pure Python implementations.

This benchmark:
- Demonstrates significant performance differences when computing primitive elements for high-degree field extensions
- Can run standalone or via ASV framework
- Includes profiling support to identify performance bottlenecks
- Tests with configurable field degrees (N=10 creates degree 2^10 = 1024)

Pure Python: 4-8 seconds | FLINT-backed: <0.1 seconds


oscarbenjamin:
This just looks AI spam.

It isn't useful to open PRs with random AI generated things. If you want to make a useful contribution that forget about the AIs and actually try reading, understanding and using the code yourself.

PR #108, opened 2026-02-22
changed the value benchmarks: runs-on ubuntu 20.04 into ubuntu latest

SirMarcoss:
working on PR#106, the CI cannot run the benchmark due to an old setting about the ubuntu version. 
changed to ubuntu-latest from ubuntu 20.04 to fix this problem


.github/workflows/run-benchmarks.yml
vprayag2005:
```suggestion
        uses: actions/checkout@v6
```

Issue #111, opened 2026-02-23
What time we expect for slow benchmark ?


vprayag2005:
see normal benchmark 
```
asv run --config asv.conf.actions.json --bench matrices
· Fetching recent changes
· Creating environments
· Discovering benchmarks.
·· Uninstalling from virtualenv-py3.14-pip+fastcache-pip+mpmath-pip+numpy..
·· Building c3326587 <master> for virtualenv-py3.14-pip+fastcache-pip+mpmath-pip+numpy.................................................................................................................
·· Installing c3326587 <master> into virtualenv-py3.14-pip+fastcache-pip+mpmath-pip+numpy..
· Running 38 total benchmarks (2 commits * 1 environments * 19 benchmarks)
[ 0.00%] · For sympy commit c3326587 <master>:
[ 0.00%] ·· Benchmarking virtualenv-py3.14-pip+fastcache-pip+mpmath-pip+numpy
[ 1.32%] ··· Running (matrices.TimeBlockDiagonalEigenvals.time_eigenvals_22--)...................
[26.32%] ··· matrices.TimeBlockDiagonalEigenvals.time_eigenvals_22                                                      2.48±0.02ms
[27.63%] ··· matrices.TimeBlockDiagonalEigenvals.time_eigenvals_222                                                     4.09±0.05ms
[28.95%] ··· matrices.TimeBlockDiagonalEigenvals.time_eigenvals_2222                                                     4.72±0.1ms
[30.26%] ··· matrices.TimeBlockDiagonalEigenvals.time_eigenvals_33                                                       3.54±0.6ms
[31.58%] ··· matrices.TimeBlockDiagonalEigenvals.time_eigenvals_333                                                      5.80±0.2ms
[32.89%] ··· matrices.TimeDiagonalEigenvals.time_eigenvals                                                              16.2±0.09μs
[34.21%] ··· matrices.TimeMatrixExpression.time_Add                                                                     3.73±0.03μs
[35.53%] ··· matrices.TimeMatrixExpression.time_MatAdd                                                                   53.1±0.9μs
[36.84%] ··· matrices.TimeMatrixExpression.time_MatAdd_doit                                                             7.16±0.03ms
[38.16%] ··· matrices.TimeMatrixExpression.time_MatMul                                                                   70.2±0.1μs
[39.47%] ··· matrices.TimeMatrixExpression.time_MatMul_doit                                                             4.96±0.01ms
[...]

PR #125, opened 2026-03-20
Add series and limit benchmarks with cache-aware variants

yuvraj707sharma:
Includes cold-cache benchmarks that measure realistic first-call performance, since ASV runs in a loop and only reports warm-cache time. See: https://github.com/sympy/sympy_benchmarks/issues/111

#### References to other Issues or PRs
See https://github.com/sympy/sympy_benchmarks/issues/111
See https://github.com/sympy/sympy/issues/21374

Add benchmarks for series expansions and limit computations.
No series or limit benchmarks existed before in this repo.

Includes cache-aware variants (TimeLimitColdCache) that call
clear_cache() before each measurement. This addresses the concern
raised in #111 that ASV loop-based timing only reports warm-cache
results, which can be misleading for SymPy operations.

Benchmarks added:
- TimeLimitWarmCache: 4 limit benchmarks (standard)
- TimeLimitColdCache: 4 limit benchmarks with cache clearing
- TimeSeriesExpansion: 5 series benchmarks parameterized by order
- TimeSeriesComposite: 3 composed expression benchmarks

#### AI Generation Disclosure
I used AI for initial guidance on the benchmark structure
and understanding the ASV format. The final code was written by me
after studying the existing benchmarks in this repo.

### DIFF
--- a/benchmarks/series.py
+++ b/benchmarks/series.py
@@ -1,0 +1,80 @@
+from sympy import symbols,oo, limit, sin, cos, exp, log, series,sqrt
+from sympy.core.cache import clear_cache
+x = symbols('x')
+
+# --- Limit benchmarks (warm cache) ---
+class TimeLimitWarmCache:
+
+    def time_limit_1_over_x(self):
+        limit(1/x, x, oo)
+
+    def time_limit_sin_x_over_x(self):
+        limit(sin(x)/x, x, 0)
+
+    def time_limit_exp(self):
+        limit((1 + 1/x)**x, x, oo)
+
+    def time_limit_log(self):
+        limit(log(x + 1)/log(x), x, oo)
+
+
+# --- Limit benchmarks (cold cache) ---
+class TimeLimitColdCache:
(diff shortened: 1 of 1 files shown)


mengChengxi:
Hello， @yuvraj707sharma 
I think this is a useful attempt to make the cache effect more visible, but I have a concern about what is actually being measured here.

Right now `clear_cache()` is called inside each `time_*` method, so the benchmark time includes both `clear_cache()` and the actual SymPy operation. That means the reported number is not strictly the first-call cost of `limit(...)` itself.

On the other hand, if `clear_cache()` were moved to `setup()`, ASV would still run the benchmark function in an inner loop, so only the first iteration would really be cold and the later ones would already be warm again.

So this seems to expose a deeper limitation of expressing true cold-cache benchmarks in ASV, rather than just a small implementation detail. It still looks useful as a “cache-aware variant”, but I am not sure it can be interpreted as a strict first-call benchmark in its current form.

PR #129, opened 2026-07-21
[GSoC] Add benchmarks for manualintegrate

salastro:
#### References to other Issues or PRs

https://github.com/sympy/sympy/issues/29862

Adds a stress-benchmark suite for `manualintegrate`, following the existing ASV-style pattern used by `benchmarks/integrate.py`'s `TimeIntegrationRisch0x` classes.

The suite adds 10 benchmark classes covering integrands that are heavier/more adversarial than the corresponding cases in `sympy/integrals/tests/test_manual.py`, targeting distinct parts of the manual-integration rule dispatcher:

- polynomial expansion over many terms
- long integration-by-parts chains (`x*exp(k*x)` sums)
- cyclic-parts IBP (`x**n*cos(x)*exp(x/4)`)
- high trig powers (sin/cos/tan/sec/cot/csc)
- symbolic (non-numeric) quadratic denominators/radicals, forcing full `Piecewise` sign-case branching
- special-function products (Ei/Si/Ci/Shi/Chi/erf/li) via nested IBP
- `exp(x)*trig(x**2)**n` (substitution + Fresnel/erf branching)
- all 8 orthogonal polynomial families at symbolic degree `n`
- substitution-search stress cases (multiple candidate substitutions)
- nested/fractional-linear square roots

Each class's `time_*` method only performs the `manualintegrate` call; setup/verification (differentiating the result and comparing back to the integrand) is done in `setup`/`teardown` so it isn't included in the timed region.

#### AI Generation Disclosure

Used Claude Sonnet 5.

### DIFF
--- a/benchmarks/manualintegrate.py
+++ b/benchmarks/manualintegrate.py
@@ -121,6 +121,11 @@ class TimeManualIntegrateSymbolicQuadratic:
 
 
 class TimeManualIntegrateSpecialFunction:
+    # sympy master is currently much slower than 1.14 on these integrands
+    # (Ei(x)*Si(x) alone exceeds asv's default 60 second timeout on CI), so
+    # allow enough time for asv's warmup and repeat calls to complete.
+    timeout = 600
+
     params = [
         Ei(x) * Si(x),
         Ei(x) * Shi(x),
--- a/benchmarks/manualintegrate.py
+++ b/benchmarks/manualintegrate.py
@@ -1,0 +1,216 @@
+
+"""
+Benchmarks for sympy.integrals.manualintegrate.manualintegrate.
+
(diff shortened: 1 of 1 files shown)


benchmarks/manualintegrate.py
```diff
@@ -121,6 +121,11 @@ def time_quadratic_power_reduction(self):
 
 
 class TimeManualIntegrateSpecialFunction:
+    # sympy master is currently much slower than 1.14 on these integrands
+    # (Ei(x)*Si(x) alone exceeds asv's default 60 second timeout on CI), so
+    # allow enough time for asv's warmup and repeat calls to complete.
+    timeout = 600
```
asmeurer, line 127: Is this a reasonable timeout or should we keep the benchmarks even faster? @oscarbenjamin

PR #130, opened 2026-08-03
assumptions: LRA benchmark

JosephMehdiyev:
There are glue codes in the benchmark that I personally needed in my usecase, I will remove them, but for now I will create this PR so I don't forget about this. It should not affect anything though

There's also small SMT-Lib benchmarking, not sure if this is right place to do this

Python hash seed should be set 0.
`asv continuous master bactracking -b "^assumptions\.lra`

#### AI Generation Disclosure

Claude Opus 5.0 was heavily used on some parts of the benchmarking


JosephMehdiyev:
I am worried that not setting Python hash seed to a constant may be problematic as it greatly changes the speed of the things. I personally set the hash seed to zero locally but I did not push these changes  here.

